PR #1, opened 2015-02-23
(CFACT-227) Move locale and catch to leatherman

branan:
These are a bit lighter-weight than the logging library, so are a
quicker start for getting it all put together.

This can be tested against https://github.com/branan/cfacter/tree/imp/master/CFACT-227-leatherman

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,32 @@
+cmake_minimum_required(VERSION 2.8.12)
+project(LEATHERMAN)
+include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/leatherman.cmake")
+
+# If we're the top-level project, we want to ensure the build type is
+# sane, and flag ourselves as such for later checks
+if ("${CMAKE_PROJECT_NAME}" STREQUAL "${PROJECT_NAME}")
+    if (NOT CMAKE_BUILD_TYPE)
+        message(STATUS "Defaulting to a release build.")
+	set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel." FORCE)
+    endif()
+
+    set(LEATHERMAN_TOPLEVEL TRUE)
+else()
+    set(LEATHERMAN_TOPLEVEL FALSE)
+endif()
+
+# If we're the top-level project enable everything by default
+defoption(LEATHERMAN_DEFAULT_ENABLE "Should Leatherman libraries all be built by default" ${LEATHERMAN_TOPLEVEL})
+defoption(LEATHERMAN_DEBUG "Enable verbose logging messages from leatherman macros" FALSE)
+defoption(LEATHERMAN_ENABLE_TESTING "Build the leatherman test binary" ${LEATHERMAN_DEFAULT_ENABLE})
+
+add_leatherman_dir(catch EXCLUDE_FROM_VARS)
+add_leatherman_dir(locale)
+
+export_var(LEATHERMAN_INCLUDE_DIRS)
+export_var(LEATHERMAN_LIBRARIES)
+
+if(LEATHERMAN_ENABLE_TESTING)
+    enable_testing()
+    add_subdirectory(tests)
+endif()
--- a/README.md
+++ b/README.md
@@ -1,0 +1,84 @@
+# Leatherman - a C++ toolkit
+
+## Usage
+
+Leatherman is intended to be included as a git submodule and added as
+a CMake subdirectory. Consider the following:
+
+    CMakeLists.txt
+    lib/
+        CMakeLists.txt
+    vendor/
+        leatherman/
+
+In this setup, your CMakeLists.txt would need to contain the following:
+
+    ...
+    add_subdirectory(vendor/leatherman)
+    ...
+
+### Enabling leatherman components
+
+Leatherman is broken up into a number of small, focused
+libraries. Each of these can be individually enabled with
+`LEATHERMAN_ENABLE_<LIBRARY>`. Any libraries not explicitly enabled
+will not be built or available to the containing project.
+
+    ...
+    set(LEATHERMAN_ENABLE_LOCALE TRUE)
+    add_subdirectory(vendor/leatherman)
+    ...
+
+### Variables set by leatherman
+
+Leatherman sets two top-level CMake variables:
+
+* `LEATHERMAN_INCLUDE_DIRS` The include paths of all enabled
+  leatherman libraries
+* `LEATHERMAN_LIBRARIES` The library names of all enabled leatherman
+  libraries, as well as their dependencies.
+
+In addition, each enabled library sets a number of library-specific
+variables:
+
+* `LEATHERMAN_<LIBRARY>_INCLUDE` The include directory or directories
+  for the given leatherman library.
+* `LEATHERMAN_<LIBRARY>_LIB` The library name as used by CMake. In the
+  case of header-only leatherman libraries, this will be set to the
+  empty string.
+* `LEATHERMAN_<LIBRARY>_DEPS` Any dependency libraries needed by the
+  given library. This could include other leatherman libraries or
+  3rd-party libraries found via CMake.
+* `LEATHERMAN_<LIBRARY>_LIBS` The contents of both
+  `LEATHERMAN_<LIBRARY>_LIB` and `LEATHERMAN_<LIBRARY>_DEPS`
+
+## Extending leatherman
+
+Adding a new library to leatherman is easy!
+
+* Add a new subdirectory with the name of your library
+* Add an appropriate `add_leatherman_dir` invocation to the top-level
+  CMakeLists.txt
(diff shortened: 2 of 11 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+# This macro gives us a pile of common variable names and behaviors
+# for leatherman libraries that we can work with
+#
+# * all the common LEATHERMAN_<LIBNAME>_{LIB,LIBS,INCLUDE,DEPS}
+#   variables for integrating individual leatherman libraries into a
+#   top-level project
+# * automatic test integration with the top-level test executable
+# * variables passed into each subdir for common naming conventions
+# * options to enable/disable individual libraries
+macro(add_leatherman_dir dir)
```
MikaelSmith: Can we pull everything about add_leatherman_dir into a separate CMake module?

catch/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+# Just point the user at our vendored catch. Nothing else to do here
```
MikaelSmith: What's the value in this redirect?

CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+    set(${lib_var} "${libname}" PARENT_SCOPE)
+endmacro()
+
+macro(add_leatherman_test)
+    add_library(${testlibname} STATIC ${ARGV})
+    set(${testlib_var} "${testlibname}" PARENT_SCOPE)
+endmacro()
+
+# This macro gives us a pile of common variable names and behaviors
+# for leatherman libraries that we can work with
```
MikaelSmith: The comments should point out the pattern used for enabling libraries.

CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+
+add_leatherman_dir(catch EXCLUDE_FROM_VARS)
+add_leatherman_dir(locale)
+
+export_var(LEATHERMAN_INCLUDE_DIRS)
+export_var(LEATHERMAN_LIBRARIES)
+
+if(LEATHERMAN_ENABLE_TESTING)
+    enable_testing()
+    add_subdirectory(tests)
```
MikaelSmith: I kind of like having tests in each individual library, so it's obvious which libraries are covered.

MikaelSmith:
I don't know if specifying install files makes more sense in this project or in the projects that use it. I lean towards leaving it out here until we see a reason to change it.

MikaelSmith:
We should add a README on how to use and include libraries.

catch/CMakeLists.txt
branan: Avoiding a special-case around catch in the top-level CMake file

CMakeLists.txt
branan: That's the intended pattern. This is just for the test executable. I'll rename the directory and clarify that with a comment down here.

CMakeLists.txt
branan: definitely

catch/CMakeLists.txt
MikaelSmith: Alright, I wonder about having the actual source over in vendor/catch then. We could point out that it's vendored in some other way.

MikaelSmith: I see why this makes sense (we may eventually have vendored libraries that we don't plan to export), and the vendor dir pattern is a useful identifier. So this is fine.

CMakeLists.txt
MikaelSmith: Oh I see, each library adds a test lib.

branan:
@MikaelSmith it's probably worth having a way to build this standalone (as a pile of static libraries and headers), but I don't see it as needed for the initial conversion.

README is in-progress.

branan:
@MikaelSmith big update!

* Added README with usage instructions as well as a basic primer on adding a library
* Moved all the macros into their own `.cmake` file
* Cleaned up macro documentation

README.md
```diff
@@ -0,0 +1,84 @@
...
+
+* `LEATHERMAN_INCLUDE_DIRS` The include paths of all enabled
+  leatherman libraries
+* `LEATHERMAN_LIBRARIES` The library names of all enabled leatherman
+  libraries, as well as their dependencies.
+
+In addition, each enabled library sets a number of library-specific
+variables:
+
+* `LEATHERMAN_<LIBRARY>_INCLUDE` The include directoryor directories
```
MikaelSmith: Need a space in `directoryor`.

MikaelSmith:
This looks great. :+1:

MikaelSmith:
Is this still WIP? Looks ready to merge to me.

cmake/leatherman.cmake
```diff
@@ -0,0 +1,163 @@
...
+	    debug("Excluding values for ${id_upper} from common vars")
+	endif()
+    endif()
+endmacro(add_leatherman_dir)
+
+####
+# Helper Macros
+####
+
+# Usage: debug("Somethign cool is happening")
```
kylog: 'Something'

kylog:
:+1: 

Great work!

kylog:
@parisiale @ploubser @richardc thoughts?

ploubser:
Sorry, commented on #2 This looks really cool. :+1:

richardc:
Seems OK, looking forward to the boost-style `Include(leatherman logging catch)` style of declaring which modules are wanted.

parisiale:
Leatherman looks great.
If we agree, I can make a PR for adding the thread ID and source line to the logging format; here's how it would look like: http://git.io/AFV6.

branan:
@MikaelSmith @kylog Typos addressed. I think this is ready to be merged now

dvhwgumby:
Sorry to add to a four year old open issue, but I cannot use the Leatherman::locale component in CMake.  `find_package(Leatherman REQUIRED locale util)` works, and I can build my code by putting `${LEATHERMAN_LIBRARIES}` in the appropriate `target_link_libraries` entries.

But if I try to do things the "correct" way (by putting `Leatherman::locale` in `target_link_libraries`, CMake complains that the target was not found.  My CMake-fu is inadequate to debug this.

I pulled on this thread while beginning to work on a pull request for cpp-hocon issue 103 (not sure how to link references across github projects).

MikaelSmith:
I guess I'm not sure why you say putting `Leatherman::local` in `target_link_libraries` is the correct way? With CMake, you can add something like `message("Libs: ${LEATHERMAN_LIBRARIES}` after `include(leatherman)` to see how the variable's initialized. For cpp-hocon it appears to be `leatherman_util;leatherman_locale;/usr/local/lib/libboost_locale-mt.dylib;/usr/local/lib/libboost_system-mt.dylib;/usr/local/lib/libboost_date_time-mt.dylib;/usr/local/lib/libboost_chrono-mt.dylib;leatherman_util`. So try adding `leatherman_locale`; you'll also need to manually add any dependencies like boost locale.

dvhwgumby:
Indeed, manipulating the various includes are what I do now for Leatherman and hocon-cpp.

However CMake doesn't want users mucking with those files, especially since (who knows?) perhaps some build systems don't use them (all the ones **I** use do, but still). 

target libraries & module names are supposed to manage both includes and library files (.a, .dylib, .so, whatever).  They want us to let CMake do the heavy lifting.  For a long time I thought the poorl named `target_link_libraries` was for `-l` and `-L` but really it is "these are the third party libraries depended on -- make sure cpp can find the headers and ld can find the library objects"

Thus the preferred, and simpler mode is the user doing the merely following:

```
foo.cc:
  #include <boost/locale>
  ....

*CMakeLists.txt:*
  find_package(Boost REQUIRED COMPONENTS locale)

  add_executable(foo foo.cc)
  target_link_libraries (foo Boost::locale)
```

And if I used Leatherman locale:

```
foo.cc:
  #include <leatherman/locale>
  ....

CMakeLists.txt:

  find_package(Leatherman REQUIRED COMPONENTS locale)

  add_executable(foo foo.cc)
  target_link_libraries (foo Leatherman::locale)
```

And let Leatherman manage its own need for Boost etc transparently to the user.  

This is what I wanted to do in a pull request for HOCON but I couldn't get Leatherman to DTRT.  There's enough going on in the Leatherman .cmake files that I'm afraid to break them.

Eventually I will want to compile with VS, which I know nothing about, so I'm eager to be able to tell CMake everything I can at as high a level as I can.

MikaelSmith:
That seems like a pattern that may have been undocumented when we wrote the Leatherman CMake files (circa CMake 3.2). So I'm not familiar with it, and don't know off the top of my head why Leatherman wouldn't support it. It's probable Leatherman needs some work to let CMake manage inherited dependencies on targets.

dvhwgumby:
Yeah, Cmake's documentation, while surprisingly comprehensive, still manages to be horrible.

I don't mean to encourage gratuitous meta work on your company's CMake infrastructure.  I'm grateful you guys ship module support in lib/cmake/leatherman at all.

If you guys ever do get around to updating the leatherman CMake module definition I'll do that for hocon and send a PR.

PR #6, opened 2015-03-04
(maint) Restore comment about UNICODE on Windows

MikaelSmith:
We force use of Unicode on Windows, following the pattern defined at
http://utf8everywhere.org/. Keep the comment from cfacter pointing out
that we're making that choice.


branan:
@MikaelSmith want to add the UTF8everywhere link to that comment? (following the "why not what" principal)

MikaelSmith:
Done.

PR #7, opened 2015-03-05
(CFACT-235) Clean up handling of defoptions for use by containing projects

branan:
The defoptions that were intended to be shared with containing
projects had been tossed at the top of `cflags.cmake`. It turns out we
really want to think about these as separate things, so they've been
split out.

Also now we only pull in cflags and options if we're top-level. When
leatherman is embedded we want to rely on the calling project to deal
with this stuff.

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -1,0 +1,6 @@
+include(leatherman)
+defoption(COVERALLS "Generate code coverage using Coveralls.io" OFF)
+defoption(BOOST_STATIC "Use Boost's static libraries" OFF)
+
+# Map our boost option to the for-realsies one
+set(Boost_USE_STATIC_LIBS ${BOOST_STATIC})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,10 +3,6 @@ project(LEATHERMAN)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 include(internal)
-include(leatherman)
-include(cflags)
-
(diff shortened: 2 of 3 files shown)


MikaelSmith:
If the conditional using BOOST_STATIC stays in cflags.cmake, then we'll still have to include it before finding Boost libraries.

We should move use of BOOST_STATIC to options.cmake. Eventually we may want to just use Boost_USE_STATIC_LIBS and remove BOOST_STATIC.

branan:
Whoops, you're right

branan:
@MikaelSmith try this one on for size

cmake/options.cmake
```diff
@@ -0,0 +1,6 @@
+include(leatherman)
+defoption(COVERALLS "Generate code coverage using Coveralls.io" OFF)
+defoption(BOOST_STATIC "Use Boost's static libraries" OFF)
+
+# Map our boot option to the for-realsies one
```
MikaelSmith: `boot option`?

branan: Clearly I am incapable of typing

branan: I kinda like "boot option" >_>

branan:
Typo fixed

branan:
I really want to see this pass appveyor... I get super nervous about windows every time I change compilation flags

MikaelSmith:
It already has passed, twice.

branan:
Thrice :)

Appveyor is fast today

branan:
Travis, on the other hand...

I killed the runs for earlier versions of this code, so hopefully we'll get a result

branan:
@MikaelSmith we're green

PR #8, opened 2015-03-17
Add .cpp to cpplint extensions

ploubser:


MikaelSmith:
We're still generally following commit message guidelines. Please prefix this with (maint).

ploubser:
Updated

PR #10, opened 2015-03-20
(CFACT-265) Disable global locale setting

MikaelSmith:
Cthun uses stringstream for conversion of port numbers to string,
that's affected by setting locale globally - it introduces comma
separators for 4-digit integers. Separate setting locale for specific
uses out of initializing logging, so it can be incrementally enabled.

### DIFF
--- a/locale/CMakeLists.txt
+++ b/locale/CMakeLists.txt
@@ -1,14 +1,13 @@
 if(WIN32)
     set(LEATHERMAN_LOCALE_SRCS "src/windows/locale.cc")
-    set(PLATFORM_COMPONENTS locale)
+    find_package(Boost 1.54 REQUIRED COMPONENTS locale)
 else()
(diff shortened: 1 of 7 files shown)


MikaelSmith:
Passes manual testing with cfacter.

parisiale:
This fixes the issue we had with cthun-agent :+1:

branan:
Looks sane to me.

We'll need a separate PR against facter to initialize the filesystem locale, correct?

MikaelSmith:
I initialize the filesystem locale as part of logging, because logging uses filesystem. Not the best idea to rely on that, but at the same time logging should own it's functionality.

Maybe logging setup should have an argument to specify the locale. Globals are terrible, and the filesystem imbue is problematic.

MikaelSmith:
Updated setup_logging to take an optional locale argument, and removes setting the boost::filesystem locale. We aren't using file logs yet, and the caller should have ownership over global settings like boost::filesystem.

PR #13, opened 2015-04-13
(LTH-1) Support 'make install' in leatherman

branan:
This adds the frameworks necessary to be able to install leatherman, as well as load it with the normal CMake `find_package` facilities. A companion PR to facter demonstrates the usage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,9 @@ endif()
 defoption(LEATHERMAN_DEFAULT_ENABLE "Should Leatherman libraries all be built by default" ${LEATHERMAN_TOPLEVEL})
 defoption(LEATHERMAN_DEBUG "Enable verbose logging messages from leatherman macros" FALSE)
 defoption(LEATHERMAN_ENABLE_TESTING "Build the leatherman test binary" ${LEATHERMAN_DEFAULT_ENABLE})
+defoption(LEATHERMAN_INSTALL "Install the leatherman libraries and headers" ${LEATHERMAN_DEFAULT_ENABLE})
+
+set(BUILDING_LEATHERMAN TRUE)
 
 #As with most things, we rely on the containing project to set up the
 #common flags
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -58,3 +61,19 @@ if(LEATHERMAN_ENABLE_TESTING)
...
+    set(CMAKE_FILES
+        cmake/cflags.cmake
+        cmake/leatherman.cmake
+        cmake/options.cmake
+        cmake/leatherman_config.cmake
+    )
+    install(FILES ${CMAKE_FILES} DESTINATION "lib/cmake/leatherman/cmake/")
+
+    configure_file(LeathermanConfig.cmake.in "${PROJECT_BINARY_DIR}/LeathermanConfig.cmake" @ONLY)
+    message(STATUS ${PROJECT_BINARY_DIR})
```
peterhuene: Debugging message or intentional?  Guessing the former given it's just a path being output.

branan: Whoops, that was a debugging message. Will remove.

CMakeLists.txt
```diff
@@ -21,6 +21,9 @@ endif()
 defoption(LEATHERMAN_DEFAULT_ENABLE "Should Leatherman libraries all be built by default" ${LEATHERMAN_TOPLEVEL})
 defoption(LEATHERMAN_DEBUG "Enable verbose logging messages from leatherman macros" FALSE)
 defoption(LEATHERMAN_ENABLE_TESTING "Build the leatherman test binary" ${LEATHERMAN_DEFAULT_ENABLE})
+defoption(LEATHERMAN_INSTALL "Install the leatherman libraries and headers" ${LEATHERMAN_DEFAULT_ENABLE})
+
+set(BUILDING_LEATHERMAN TRUE)
```
MikaelSmith: When is this ever false?

MikaelSmith: Oh I see, it's so all the library cmake files can be re-used for configuration.

MikaelSmith:
Looks good; I'm going to play with installing it a bit.

MikaelSmith:
:+1:

CMakeLists.txt
branan: Yup. I'm going to update the commit message and the README with more info on how/why this is the case.

branan:
README updated. I don't think there are any other changes I want to make to this, so once the README is +1'd we're good to go.

The only change in the updated base commit was the removal of the debug message Peter found.

README.md
```diff
@@ -52,6 +87,24 @@ variables:
 * `LEATHERMAN_<LIBRARY>_LIBS` The contents of both
   `LEATHERMAN_<LIBRARY>_LIB` and `LEATHERMAN_<LIBRARY>_DEPS`
 
+### CMake Helpers Provided by Leatherman
+
+In addition to the C++ library components, Leatherman provides a few
+CMake helpers. These will be automatically added to your
+`CMAKE_MODULE_PATH` when `find_package` is processed.
```
branan: I'm kind of tempted to just include `options`, `cflags`, and `leatherman` in the Finder and obviate the need to explicitly include them. I can't think of a situation where we *wouldn't* want them to be there.

My only concern with that is certain things aren't well-namespaced. I may hold off on that until a separate PR where I can rename all the macros to have a `leatherman_` or `lth_` prefix.

branan:
@MikaelSmith Mentioned Boost in a new dependencies section, and added a note in the "Using Leatherman" section about `CMAKE_PREFIX_PATH`

MikaelSmith:
Typo in the commit message.

branan:
@MikaelSmith annnnd fixed

PR #14, opened 2015-04-16
(LTH-6) Add the source line number to logging messages

parisiale:
Adding the source line number by using the __LINE__ macro.
Updating logging unit tests.

### DIFF
--- a/logging/src/logging.cc
+++ b/logging/src/logging.cc
@@ -53,7 +53,7 @@ namespace leatherman { namespace logging {
                 << expr::format_date_time<boost::posix_time::ptime>("TimeStamp", "%Y-%m-%d %H:%M:%S.%f")
                 << " " << left << setfill(' ') << setw(5) << leatherman::logging::log_level_attr
                 << " " << leatherman::logging::namespace_attr
-                << " - " << expr::smessage);
+                << ":" << expr::smessage);
 
         boost::log::add_common_attributes();
 
@@ -139,12 +139,12 @@ namespace leatherman { namespace logging {
         return g_colorize ? reset : none;
     }
 
-    void log(const string &logger, log_level level, boost::format& message)
+    void log(const string &logger, log_level level, int line_num, boost::format& message)
     {
-        log(logger, level, message.str());
+        log(logger, level, line_num, message.str());
     }
 
-    void log(const string &logger, log_level level, string const& message)
+    void log(const string &logger, log_level level, int line_num, string const& message)
     {
         if (level >= log_level::error) {
             g_error_logged = true;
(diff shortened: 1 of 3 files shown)


parisiale:
Tested on cthun-agent, on OS X.
Ex:

2015-04-16 23:36:00.187654 TRACE puppetlabs.cthun_agent.connection:268 - WebSocket TLS initialization event

parisiale:
Created LTH-6.

MikaelSmith:
The commit message should be updated to include the ticket.

The number is low-profile, so I'm not particularly against it, but in Facter we don't really have per-file signatures anymore, so I'd be tempted to have a way to disable it via another macro.

logging/src/logging.cc
```diff
@@ -53,7 +53,7 @@ namespace leatherman { namespace logging {
                 << expr::format_date_time<boost::posix_time::ptime>("TimeStamp", "%Y-%m-%d %H:%M:%S.%f")
                 << " " << left << setfill(' ') << setw(5) << leatherman::logging::log_level_attr
                 << " " << leatherman::logging::namespace_attr
-                << " - " << expr::smessage);
+                << ":" << expr::smessage);
```
MikaelSmith: If we add a macro to disable printing the line number, I'd move the `:` formatting to line 158.

parisiale:
I've updated the title tag.
I'll look into adding a macro.

branan:
This largely looks pretty good to me. It will have some issues once #13 is merge - the big one being that I think the change to `logging.cpp` will need to be a runtime toggle, not an ifdef.

I'm happy to make those tweaks myself as part of LTH-1, though, and merge this as-is.

parisiale:
Thanks @branan 
If it will get too complicated with LTH-1, I suggest to simply drop this.

MikaelSmith:
Runtime toggle is too bad, but I bet branch-prediction would make it negligible.

branan:
@MikaelSmith yeah. Branch prediction, and/or a good LTO pass.

Also we're doing I/O here, so I don't care that much about an extra branch - should be negligable :)

MikaelSmith:
If we're going to have to make it runtime check anyway, you can drop the last two commits and we'll just merge the first.

parisiale:
I reverted the commits, to leave them in the history.
Let me know if you prefer a force push instead.

MikaelSmith:
Oh, can you force push the commits? I'd rather not keep the intermediate work, and need to fix the 1st commit to reference LTH-6.

parisiale:
Will do.

MikaelSmith:
Travis is taking so long...

I started looking ahead to what we'd do to make logging line numbers configurable per-project that will work with the LTH-1 work - @branan - and ended up with https://github.com/puppetlabs/leatherman/compare/master...MikaelSmith:maint/master/per-project-logging-line-numbers.

branan:
@MikaelSmith my idea was pretty much the same, but instead of overloading functions would pass a boolean and trust the predictor and/or LTO to sort it out for us.

Given that logging code isn't critical-path, I'm not sure we care about whether it gets optimized either way

MikaelSmith:
Ended up with a PR at https://github.com/puppetlabs/leatherman/pull/15.

PR #15, opened 2015-04-28
(LTH-8) Make logging line numbers configurable

MikaelSmith:
Makes logging line numbers a compile-time configuration option that can
be set per-project using the same compiled artifacts of Leatherman. This
is needed to support the `make install` workflow for Leatherman (LTH-1).

### DIFF
--- a/cmake/leatherman.cmake
+++ b/cmake/leatherman.cmake
@@ -9,6 +9,13 @@ macro(leatherman_logging_namespace namespace)
     add_definitions("-DLEATHERMAN_LOGGING_NAMESPACE=\"${namespace}\"")
 endmacro()
 
+# Usage: leatherman_logging_line_numbers()
+#
+# Sets the LEATHERMAN_LOGGING_LINE_NUMBERS preprocessor definition.
+macro(leatherman_logging_line_numbers)
+    add_definitions("-DLEATHERMAN_LOGGING_LINE_NUMBERS")
+endmacro()
+
 # Usage: debug("Something cool is happening")
 #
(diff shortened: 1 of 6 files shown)


branan:
This looks sane to me, and should work correctly with LTH-1 merged.

We could probably avoid the duplicate function body with a boolean parameter next to the line number param - `print_line_number` or something like that. Logging isn't exactly performance-critical, and we are already doing a bunch of allocations and IO around this code anyway.

On the other hand, it's not that much duplicated code. I don't have strong feelings, just tossing thoughts out.

I'm :+1: on this as it stands, to be clear.

MikaelSmith:
I'd be alright re-factoring it, I'm not to keen on duplication either.

branan:
If you've got the time for it, that'd be awesome.

branan:
That looks good to me (assuming we don't anticipate files with more than about 2 billion lines :wink:)

kylog:
Is `__LINE__` zero-based? That seems unintuitive. If not, should the `>= 0` logic here be `> 0`?

branan:
IIRC `__LINE__` is one-based, but this code is explicitly using -1 as a sentinel value for "don't print" in the macros. If you're using the macros like you should be, it'll either be >= 1 or -1, so both tests will work for our purposes.

kylog:
Agreed that the tests will work either way because `0` will never happen. Mine was a readability question not a functional question, i.e. specifying `>= 0` as a test for a valid line number just seems weird to me. 

But nbd here if that test makes most sense to others.

MikaelSmith:
I like changing the sentinel to 0. That allows us to use unsigned ints in the future.

PR #16, opened 2015-06-12
(LTH-11) Move DataContainer class to Leatherman

Magisus:
The DataContainer classes from cthun-client and deployer-client are extracted into Leatherman and renamed JsonContainer.

### DIFF
--- a/rapidjson/CMakeLists.txt
+++ b/rapidjson/CMakeLists.txt
@@ -1,0 +1,10 @@
+if (BUILDING_LEATHERMAN)
+    set(include_dir "${PROJECT_SOURCE_DIR}/vendor/rapidjson/include")
+    set(${include_var} ${include_dir} PARENT_SCOPE)
+
+    if(LEATHERMAN_INSTALL)
+        install(DIRECTORY "${include_dir}" DESTINATION "include/leatherman/vendor/rapidjson")
+    endif()
+else()
+    set(${include_var} "${LEATHERMAN_PREFIX}/include/leatherman/vendor/rapidjson/include")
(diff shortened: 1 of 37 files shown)


rapidjson/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+if (BUILDING_LEATHERMAN)
+    set(include_dir "${PROJECT_SOURCE_DIR}/vendor/rapidjson/include")
+    set(${include_var} ${include_dir} PARENT_SCOPE)
+
+    if(LEATHERMAN_INSTALL)
+        install(DIRECTORY "${include_dir}" DESTINATION "include/leatherman/vendor/rapidjson")
```
MikaelSmith: I think this will need to copy the `rapidjson` directory in `${include_dir}`. The `${include_var}` at line 9 will add the destination to include paths, and `install(DIRECTORY ...)` will move the named directory to the destination.

So the way this is currently written, I'd expect to find `include/leatherman/vendor/rapidjson/include/rapidjson/*`, which would require doing `include <include/rapidjson/*>` to include something. We don't have good tests for this, so nothing's likely to break if it's wrong yet.

Alternatively, you could append `include` to the path at line 9.

MikaelSmith: You should try `make install` in the leatherman project to see what you end up with.

MikaelSmith:
Travis CI is running `make cpplint` and failing.

MikaelSmith:
I believe this is rapidjson 0.11. We should add that in the commit message.

PR #19, opened 2015-06-15
(LTH-7) Fix LEATHERMAN_LIBRARIES ordering

MikaelSmith:
Without this change, Windows release builds use debug libraries because
the optimized/debug labeling CMake adds in the LIBS variables gets
reversed. Insert libraries in reverse order rather than reversing at the
end so we get the labels in the correct order.


MikaelSmith:
Looks like I might need to ensure libs don't end up in deps.

MikaelSmith:
Fixed and verified in Facter.

branan:
This looks more correct than the old version to me, and "works on my laptop". Let's see what happens

PR #22, opened 2015-06-17
(LTH-12) Extract facter Windows helpers and wrappers to Leatherman

Magisus:
Move the registry, user, process, system_error, and wmi Windows helper classes from facter to Leatherman, creating a Windows utility library for use in other projects.

### DIFF
--- a/util/inc/leatherman/util/scoped_resource.hpp
+++ b/util/inc/leatherman/util/scoped_resource.hpp
@@ -1,0 +1,130 @@
+/**
+ * @file
+ * Declares the base class for scoped resources.
+ */
+#pragma once
+
+#include <functional>
+
+namespace leatherman { namespace util {
+    /**
+     * Simple class that is used for the RAII pattern.
+     * Used to scope a resource.  When it goes out of scope, a deleter
+     * function is called to delete the resource.
+     * This type can be moved but cannot be copied.
+     * @tparam T The type of resource being scoped.
+    */
+    template<typename T>
+    struct scoped_resource
+    {
+        /**
+         * Constructs an uninitialized scoped_resource.
+         * Can be initialized via move assignment.
+         */
+        scoped_resource() :
+            _resource(),
+            _deleter(nullptr)
+        {
+        }
+
+        /**
+         * Constructs a scoped_resource.
+         * Takes ownership of the given resource.
+         * @param resource The resource to scope.
+         * @param deleter The function to call when the resource goes out of scope.
+         */
+        scoped_resource(T resource, std::function<void(T&)> deleter) :
+            _resource(std::move(resource)),
+            _deleter(deleter)
+        {
+        }
+
+        /**
+         * Prevents the scoped_resource from being copied.
+         */
+        explicit scoped_resource(scoped_resource<T> const&) = delete;
+        /**
+         * Prevents the scoped_resource from being copied.
+         * @returns Returns this scoped_resource.
+         */
+        scoped_resource& operator=(scoped_resource<T> const&) = delete;
+        /**
+         * Moves the given scoped_resource into this scoped_resource.
+         * @param other The scoped_resource to move into this scoped_resource.
+         */
+        scoped_resource(scoped_resource<T>&& other)
+        {
+            *this = std::move(other);
+        }
+
+        /**
+         * Moves the given scoped_resource into this scoped_resource.
+         * @param other The scoped_resource to move into this scoped_resource.
+         * @return Returns this scoped_resource.
+         */
+        scoped_resource& operator=(scoped_resource<T>&& other)
+        {
+            release();
+            _resource = std::move(other._resource);
+            _deleter = std::move(other._deleter);
+
+            // Ensure the deleter is in a known "empty" state; we can't rely on default move semantics for that
+            other._deleter = nullptr;
+            return *this;
+        }
+
+        /**
+         * Destructs a scoped_resource.
+         */
+        ~scoped_resource()
+        {
+            release();
+        }
+
+        /**
+         * Implicitly casts to T&.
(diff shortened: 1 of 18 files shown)


windows/inc/leatherman/windows/scoped_resource.hpp
```diff
@@ -0,0 +1,130 @@
+/**
+ * @file
+ * Declares the base class for scoped resources.
+ */
+#pragma once
+
+#include <functional>
+
+namespace leatherman { namespace windows {
+    /**
```
kylog: Is this windows-specific?

windows/inc/leatherman/windows/user.hpp
```diff
@@ -0,0 +1,30 @@
...
+     * tasks restricted to Administrators.
+     * @return True if the current process has Administrator privileges, otherwise false.
+     */
+    bool is_admin();
+
+    /**
+     * Query token membership to determine whether the current user is a member of the Administrators group.
+     * @return True if user is an Administrator, otherwise false.
+     */
+    bool check_token_membership();
```
kylog: Today I only see this used in the `is_admin` implementation. Is it of general utility? If not, should we not expose it? 

Either way, is there a better name for the function? Like `in_admin_group`? Otoh if `check_token_membership` will be intuitively obvious to someone doing windows work, then just ignore me :)

Magisus:
No, in facter it was outside of the windows stuff. And actually, apparently the file that included it (wmi.cc) doesn't even use it. So I guess it can be removed from this library? I left the original in facter/inc/facter/util, because I figured it was used elsewhere.

windows/inc/leatherman/windows/wmi.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+        /**
+         * Identifier for the WMI property OtherTypeDescription
+         */
+        constexpr static char const* othertypedescription = "OtherTypeDescription";
+
+        /**
+         * Case-insensitive string comparison.
+         */
+        struct ciless : std::binary_function<std::string, std::string, bool>
```
kylog: Extract from the wmi namespace?

windows/inc/leatherman/windows/scoped_resource.hpp
kylog: Yeah, I think `scoped_resource` isn't windows specific, but of broad utility. If so, let's see if we can pull up into a non-windows util header somewhere.

windows/inc/leatherman/windows/user.hpp
MikaelSmith: These were extracted from https://github.com/puppetlabs/puppet/blob/003d2b73b4730953092bb7e8b7bc723d2210c999/lib/puppet/util/windows/user.rb, and will likely be needed in a future native client. Both are marked module functions, so I suspect they would be useful elsewhere.

util/inc/leatherman/util/scoped_resource.hpp
```diff
@@ -0,0 +1,130 @@
+/**
+ * @file
+ * Declares the base class for scoped resources.
+ */
+#pragma once
+
+#include <functional>
+
+namespace leatherman { namespace windows {
```
MikaelSmith: This probably shouldn't be in the windows namespace.

util/inc/leatherman/util/scoped_resource.hpp
```diff
@@ -0,0 +1,130 @@
...
+        std::function<void(T&)> _deleter;
+
+     private:
+        void* operator new(size_t) = delete;
+        void operator delete(void*) = delete;
+        void* operator new[](size_t) = delete;
+        void operator delete[](void* ptr) = delete;
+    };
+
+}}  // namespace leatherman::windows
```
MikaelSmith: Same note about windows here.

util/inc/leatherman/util/strings.hpp
```diff
@@ -0,0 +1,30 @@
+/**
+ * @file
+ * Declares utility functions for dealing with strings.
+ */
+#pragma once
+
+#include <boost/algorithm/string/predicate.hpp>
```
MikaelSmith: I think you can drop `boost/range/iterator_range.hpp`.

windows/inc/leatherman/windows/registry.hpp
```diff
@@ -0,0 +1,56 @@
...
+         * @param hkey The registry key handle.
+         * @param subkey The name of the registry key.
+         * @param value The name of the registry value.
+         * @return An array of string values corresponding to the REG_MULTI_SZ type.
+         *         Returns an empty vector if the value doesn't exist or isn't a composite string type.
+         */
+        std::vector<std::string> get_registry_strings(HKEY hkey, std::string const& subkey, std::string const& value);
+    }
+
+}}  // leatherman::windows
```
MikaelSmith: completely trivial, but the others use `namespace leatherman::windows`.

windows/inc/leatherman/windows/wmi.hpp
```diff
@@ -0,0 +1,190 @@
+/**
+ * @file
+ * Declares utility functions for interacting with Windows Management Instrumentation.
+ */
+#pragma once
+
+#include <leatherman/util/scoped_resource.hpp>
+#include <leatherman/util/strings.hpp>
+#include <boost/algorithm/string/predicate.hpp>
```
MikaelSmith: Both boost/algorithm/string includes aren't needed here.

MikaelSmith:
We should add a README note that the `windows` module requires logging to be setup.

MikaelSmith:
I'd like to eventually have stand-alone tests for these utilities, so filed [LTH-23](https://tickets.puppetlabs.com/browse/LTH-23).

PR #26, opened 2015-06-19
(maint) Fix compiling logging tests with Clang

MikaelSmith:
### (maint) Fix compiling logging tests with Clang
Compiling with clang fails, move operator declarations to the boost
namespace because that seems to work.

I'm not really sure why, seems related to https://en.wikipedia.org/wiki/Argument-dependent_name_lookup.

### (maint) Fix testing in Clang
With GCC, somehow the test worked even though padding should've
introduced multiple spaces. Clang showed the error in the test.

Add handling for multiple spaces following the log level string.

### DIFF
--- a/logging/tests/logging.cc
+++ b/logging/tests/logging.cc
@@ -1,17 +1,21 @@
 #include "logging.hpp"
 #include <boost/nowide/iostream.hpp>
 
-namespace leatherman { namespace test {
+namespace boost {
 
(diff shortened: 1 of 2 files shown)


MikaelSmith:
@branan

branan:
Would be cool to know *why* the GCC version of the test passes - behavioral differences at runtime scare me - but I can hit the button on this for now.

PR #28, opened 2015-06-22
(LTH-17) Extract various file and directory utility classes to LTH

Magisus:
Extracts the file utility classes from facter, cthun-agent, and Pegasus to Leatherman.


file_util/inc/leatherman/file_util/directory.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include <string>
+#include <functional>
+
+namespace leatherman { namespace file_util {
+
+    /**
+     * Contains utility functions for enumerating directories.
+     */
+    struct __attribute__((visibility("default"))) directory
```
Magisus: In facter, these methods were bundled in a struct like this. Is that something we want to continue doing? Why was it done this way? (Still getting up to speed on the subtleties of C++)

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -0,0 +1,117 @@
...
+    /// Generic file error class.
+    class file_error : public std::runtime_error {
+    public:
+        explicit file_error(std::string const& msg) : std::runtime_error(msg) {}
+    };
+
+    /**
+     * Contains utility functions for reading data from files.
+     */
+    struct __attribute__((visibility("default"))) file
```
Magisus: Same question as with the directory file: why bundled in a struct? Should it stay this way?

Magisus:
This currently does not build on Windows, due to not being able to find wordexp.h, which is used in the methods extracted from cthun-agent. I assume we want this library to be available on Windows, so what is the best way to resolve this issue?

file_util/tests/test.hpp
```diff
@@ -0,0 +1,9 @@
+#ifdef FAIL
```
MikaelSmith: This whole file seems unused.

Magisus: Yeah, I guess it just needs the catch include. Moved that into the test class itself.

file_util/inc/leatherman/file_util/directory.hpp
```diff
@@ -0,0 +1,34 @@
+/**
+ * @file
+ * Declares utility functions for enumerating directories.
+ */
+#pragma once
+
+#include <string>
+#include <functional>
+
+namespace leatherman { namespace file_util {
```
MikaelSmith: Do we want to preserve distinct namespaces for file and directory? There aren't any conflicts, so seems safe enough to put them all in one namespace.

file_util/inc/leatherman/file_util/directory.hpp
```diff
@@ -0,0 +1,34 @@
...
+     * @param callback The callback to invoke when a matching subdirectory is found.
+     * @param pattern The pattern to filter the subdirectory names by.  If empty, all subdirectories are passed.
+     */
+    void each_subdirectory(std::string const& path, std::function<bool(std::string const&)> callback, std::string const& pattern = {});
+
+    /**
+     * Create a directory.
+     * @return Returns true on success, false on failure.
+     */
+    bool createDirectory(const std::string& dirname);
```
MikaelSmith: Let's use snake_case instead of camelCase. It matches styles already used in our community (Ruby).

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -0,0 +1,111 @@
...
+ * Declares utility functions for reading data from files.
+ */
+#pragma once
+
+#include <boost/filesystem.hpp>
+
+#include <string>
+#include <stdexcept>
+#include <functional>
+#include <wordexp.h>
```
MikaelSmith: Not needed in the header file, this could be moved to the source file.

file_util/inc/leatherman/file_util/regex.hpp
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Helper function for resolving variadic arguments to re_search.
+     * @tparam Text The type of the text to search.
+     * @param txt The text to search.
+     * @param what The pattern to search the text with.
+     * @param depth The current argument depth.
+     * @return Returns true if the match group was found or false if it was not.
+     */
+    template <typename Text>
+    inline bool re_search_helper(Text &txt, const boost::smatch &what, size_t depth)
```
MikaelSmith: I think this is more appropriate in the leatherman/util library.

file_util/src/directory.cc
```diff
@@ -0,0 +1,71 @@
...
+            if (re_search(it->path().filename().string(), regex)) {
+                if (!callback(it->path().string())) {
+                    break;
+                }
+            }
+        }
+    }
+
+    bool createDirectory(const std::string& dirname) {
+        boost::filesystem::path dir(dirname);
```
MikaelSmith: This helper seems unnecessary. `boost::filesystem::create_directory` can take a `string` argument directly, so you should be able to replace `createDirectory` with `boost::filesystem::create_directory`.

file_util/src/file.cc
```diff
@@ -0,0 +1,162 @@
...
+            return "";
+        }
+
+        // Get the expanded text and free the memory
+        std::string expanded_txt { result.we_wordv[0] };
+        wordfree(&result);
+        return expanded_txt;
+    }
+
+    bool fileExists(const std::string& file_path) {
```
MikaelSmith: This helper really checks if the file exists and is readable by the current process. If that's what's needed, we should rename it to something like `file_readable`.

file_util/src/file.cc
```diff
@@ -0,0 +1,162 @@
...
+            std::ifstream file_stream { file_path };
+            exists = file_stream.good();
+            file_stream.close();
+        }
+        return exists;
+    }
+
+    void removeFile(const std::string& file_path) {
+        if (file_util::fileExists(file_path)) {
+            if (std::remove(file_path.c_str()) != 0) {
```
MikaelSmith: Also has a problem with unicode on Windows, and could be replaced directly with `boost::filesystem::remove` which throw `boost::system::error_code`.

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -0,0 +1,111 @@
...
+#include <fstream>
+
+namespace leatherman { namespace file_util {
+
+    /// Generic file error class.
+    class file_error : public std::runtime_error {
```
MikaelSmith: I'd lean towards replacing this with `boost::filesystem::filesystem_error` for consistency.

file_util/src/file.cc
```diff
@@ -0,0 +1,162 @@
...
+        ofs << text;
+        ofs.close();
+        rename(tmp_name.data(), file_path.data());
+    }
+
+    void writeToFile(const std::string& text, const std::string& file_path) {
+        streamToFile(text, file_path, std::ofstream::out | std::ofstream::trunc);
+    }
+
+    std::string readFileAsString(std::string path) {
```
MikaelSmith: I'm not sure this differs from the above `read` function.

file_util/src/file.cc
```diff
@@ -0,0 +1,162 @@
...
+
+    void removeFile(const std::string& file_path) {
+        if (file_util::fileExists(file_path)) {
+            if (std::remove(file_path.c_str()) != 0) {
+                throw file_error { "failed to remove " + file_path };
+            }
+        }
+    }
+
+    void streamToFile(const std::string& text,
```
MikaelSmith: I'm not sure how streamToFile and writeToFile differ, @parisiale?

file_util/src/file.cc
```diff
@@ -0,0 +1,162 @@
...
+
+        return content;
+    }
+
+    std::string tildeExpand(std::string path) {
+        if (path[0] == '~' && (path.size() == 1 || path[1] == '/')) {
+            #ifdef _WIN32
+                std::string result { getenv("USERPROFILE") };
+            #else
+                std::string result { getenv("HOME") };
```
MikaelSmith: This should use boost::nowide::getenv for unicode handling, in case the path has unicode characters.

file_util/src/file.cc
```diff
@@ -0,0 +1,162 @@
...
+
+            result.append(path.begin() + 1, path.end());
+            return result;
+        }
+        return path;
+    }
+
+    std::string shellQuote(std::string path) {
+        std::stringstream ss;
+        ss << boost::io::quoted(path);
```
MikaelSmith: What's the header file for `boost::io`?

file_util/src/file.cc
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    std::string readFileAsString(std::string path) {
+        std::string content = "";
+
+        if (!fileExists(path)) {
+            return content;
+        }
+
+        std::ifstream file { path };
```
MikaelSmith: Use `boost::nowide::ifstream`.

file_util/src/file.cc
```diff
@@ -0,0 +1,162 @@
...
+            if (std::remove(file_path.c_str()) != 0) {
+                throw file_error { "failed to remove " + file_path };
+            }
+        }
+    }
+
+    void streamToFile(const std::string& text,
+                      const std::string& file_path,
+                      std::ios_base::openmode mode) {
+        std::ofstream ofs;
```
MikaelSmith: Use `boost::nowide::ofstream`.

file_util/tests/file_utils_test.cc
```diff
@@ -0,0 +1,80 @@
...
+#endif
+
+namespace leatherman { namespace file_util {
+
+// TODO(ale): consider making file_utils.cpp and string_utils.cpp
+// consistent across cthun-agent and pegasus before writing more tests
+
+TEST_CASE("file_util::tildeExpand", "[utils]") {
+#ifdef _WIN32
+    _putenv("USERPROFILE=/testhome");
```
MikaelSmith: You could use `boost::nowide::setenv` to avoid the `#ifdef`.

file_util/src/file.cc
Magisus: This seems to be used when setting config files (see cthun-agent/lib/src/configuration.cc) and to trigger some errors (see lib/src/modules/status.cc). Is readability an important part of those tasks? Seems like maybe for the config files it is? Unless of course they would never not be readable...

MikaelSmith: Those are all reads. The easiest thing to do is change the name and use `boost::nowide::ifstream`.

file_util/src/file.cc
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    void atomic_write_to_file(const std::string &text,
+                              const std::string &file_path,
+                              std::ios_base::openmode mode = ios::binary) {
+        boost::nowide::ofstream ofs;
+        std::string tmp_name = file_path + "~";
+        ofs.open(tmp_name, mode);
+        if (!ofs.is_open()) {
+            throw std::runtime_error { "failed to open " + file_path };
```
Magisus: I was having trouble getting filesystem_error to work, as a second argument it wanted an error_code, which is an enum defined in boost::filesystem, but it didn't want to compile, most likely because I didn't chance on the right syntax. Do you have any insight?

file_util/tests/file_utils_test.cc
Magisus: Found the documentation on setenv, but I'm still unclear on how to use it here...

file_util/src/file.cc
MikaelSmith: http://www.boost.org/doc/libs/1_33_1/libs/filesystem/doc/exception.htm#Members

file_util/src/file.cc
```diff
@@ -0,0 +1,140 @@
+#include <leatherman/file_util/file.hpp>
+#include <boost/nowide/fstream.hpp>
+#include <boost/nowide/cenv.hpp>
+#include <boost/system/error_code.hpp>
+#include <sstream>
+#include <leatherman/logging/logging.hpp>
+
+#include <wordexp.h>
```
Magisus: This was added by @parisiale to the cthun-agent. It does not work on Windows. What would be a good alternative/fix/workaround?

file_util/inc/leatherman/file_util/directory.hpp
ploubser: Feel free.

file_util/src/file.cc
ploubser: @Magisus We have replaced shellExpand with https://github.com/puppetlabs/pegasus/blob/master/lib/src/file_utils.cc#L30 in Pegasus which doesn't rely on wordexp.h @parisiale This is all we need in the agent too, right?

Magisus: Looking at the tests, this seems right. I'll consolidate them.

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -0,0 +1,78 @@
...
+    void atomic_write_to_file(const std::string &text,
+                              const std::string &file_path,
+                              std::ios_base::openmode mode = std::ios::binary);
+
+    /**
+     * Expands a leading tilde to the user's home directory
+     * @return Returns the expanded path, or the original string
+     *         in case the expansion fails.
+     */
+    std::string tildeExpand(std::string path);
```
MikaelSmith: Are these being removed or renamed to use snake_case?

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Returns a shell-safe version of the path
+     */
+    std::string shellQuote(std::string path);
```
MikaelSmith: Removed or snake_case?

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -0,0 +1,78 @@
...
+        boost::filesystem::path source;
+        std::string relativeName;
+    };
+
+    using FileList = std::vector<FileCopy>;
+
+    /**
+     * Returns a set of files suitable for copying
+     */
+    FileList relativeFileList(boost::filesystem::path path);
```
MikaelSmith: Removed or snake_case?

MikaelSmith:
Are there tests we're missing? Or were they just never there?

file_util/inc/leatherman/file_util/file.hpp
Magisus: Everything left in here is staying, doing the rename last because you said you wanted a separate commit, and I wasn't 100% how best to handle that otherwise.

Magisus:
Mostly they were never there. There are a couple additional ones in cthun-agent that I might grab, but some are redundant and others are reliant on other files from cthun-agent. I can extract those files too if you think it's worth it. (see cthun-agent/uuid.hpp)

MikaelSmith:
I think we could extract those, and look at adding some new tests for the file utils that don't have any yet. I don't like the 3% drop in coverage.

file_util/src/file.cc
```diff
@@ -0,0 +1,123 @@
...
+            throw boost_file::filesystem_error { "failed to open " + file_path,
+                                                        boost_error::make_error_code(
+                                                                boost_error::io_error) };
+        }
+        ofs << text;
+        ofs.close();
+        rename(tmp_name.data(), file_path.data());
+    }
+
+    std::string tilde_expand(std::string path) {
```
Magisus: Every test of this method is failing on Windows due to a "basic_string::_S_construct null not valid" exception. Is there any way to get more information on where this exception is coming from when running the tests?

file_util/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+find_package(Boost 1.54 REQUIRED COMPONENTS filesystem)
```
MikaelSmith: Boost.Filesystem depends on Boost.System. I'd append `system` after `filesystem`, and use `${Boost_LIBRARIES}' for `add_leatherman_deps`.

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Reads the entire contents of the given file into a string.
+     * @param path The path of the file to read.
+     * @param contents The returned file contents.
+     * @return Returns true if the contents were read or false if the file is not readable.
+     */
+    bool read(std::string const& path, std::string& contents);
+
+    /**
+     *@return Returns true if the specified file exists.
```
MikaelSmith: 'exists and is can be read by the current process.'

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -0,0 +1,82 @@
...
+     */
+    bool file_readable(const std::string &file_path);
+
+    /**
+     * Writes content to file in the specified mode.If file exists,
```
MikaelSmith: Add a space between '.If'.

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -0,0 +1,82 @@
...
+    void atomic_write_to_file(const std::string &text,
+                              const std::string &file_path,
+                              std::ios_base::openmode mode = std::ios::binary);
+
+    /**
+     * Expands a leading tilde to the user's home directory
+     * @return Returns the expanded path, or the original string
+     *         in case the expansion fails.
+     */
+    std::string tilde_expand(std::string path);
```
MikaelSmith: Should add `@param path ...`.

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Returns a shell-safe version of the path
+     */
+    std::string shell_quote(std::string path);
```
MikaelSmith: Should add doxygen tags for `@param`, `@return`.

MikaelSmith:
At some point we may want to clean up doxygen comments, but right now we don't have doxygen setup.

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+    struct FileCopy {
+        boost::filesystem::path source;
+        std::string relativeName;
```
MikaelSmith: I'd also use snake_case here.

file_util/src/file.cc
MikaelSmith: Resolved.

file_util/src/file.cc
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    bool file_readable(const std::string &file_path) {
+        bool exists { false };
+        if (file_path.empty()) {
+            LOG_WARNING("file path is an empty string");
+        } else{
+            boost::system::error_code ec;
+            boost_file::file_status status = boost_file::status(file_path.c_str(), ec);
+            if (boost_file::status_known(status)) {
```
MikaelSmith: Is that sufficient to check whether it's a file (and not a directory)? I think we also need to check `exists` and `!is_directory`.

MikaelSmith:
Can we add tests for `each_line`, `read`, `each_file`, and `each_subdirectory`?

file_util/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+
+add_leatherman_deps("${Boost_FILESYSTEM_LIBRARY}")
+add_leatherman_includes("${Boost_INCLUDE_DIRS}")
+
+leatherman_dependency(nowide)
+leatherman_dependency(logging)
+leatherman_dependency(util)
+
+if (BUILDING_LEATHERMAN)
+    leatherman_logging_namespace("leatherman.logging")
```
branan: Should we be using the `leatherman.file_util` logging namespace here?

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Expands a leading tilde to the user's home directory
+     * @return Returns the expanded path, or the original string
+     *         in case the expansion fails.
+     */
+    std::string tilde_expand(std::string path);
+
+    /**
+     * Returns a shell-safe version of the path
```
whopper: Should this be `@return Returns a shell-safe version of the path`? Same for a few more functions below.

MikaelSmith: Leatherman doesn't have doxygen setup yet. That's something we should do, but I'm fine doing it in a separate ticket.

whopper: Ah cool, I'll put up a ticket so we don't forget in the future!

file_util/src/file.cc
```diff
@@ -0,0 +1,142 @@
...
+        buffer << in.rdbuf();
+        contents = buffer.str();
+        return true;
+    }
+
+    bool file_readable(const std::string &file_path) {
+        bool exists { false };
+        if (file_path.empty()) {
+            LOG_WARNING("file path is an empty string");
+        } else{
```
whopper: Nitpick: missing a space between `else` and `{`

file_util/tests/directory_utils_test.cc
```diff
@@ -0,0 +1,82 @@
+#include <catch.hpp>
+#include <leatherman/file_util/file.hpp>
+#include <leatherman/file_util/directory.hpp>
+
+#ifdef _WIN32
+#include <windows.h>
```
MikaelSmith: I don't think this include is needed.

MikaelSmith: Let's pair up on the test cases, I think there's a number of things that can be improved.

MikaelSmith:
Except for cleaning up the test cases, :+1:. I think we can squash the commits now as well.

whopper:
Yep, besides my super nickpick that you already addressed, :+1: from me :)

tests/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 find_package(Threads)
 
 include_directories(${LEATHERMAN_CATCH_INCLUDE} ${LEATHERMAN_INCLUDE_DIRS})
-add_executable(leatherman_test main.cc ${LEATHERMAN_TEST_SRCS})
+add_executable(leatherman_test main.cc ${LEATHERMAN_TEST_SRCS} ../file_util/tests/fixtures.hpp)
```
MikaelSmith: I'm not really sure this is needed.

MikaelSmith:
:+1:

Magisus:
So this test that is failing on Travis (for directory/each_file) is not failing anywhere else... what's the best way to debug this behavior?

PR #29, opened 2015-06-22
(maint) Add get_commit_string helper

MikaelSmith:
Adds a get_commit_string helper for getting a commit description string
in CMake.


MikaelSmith:
This should be followed with an update to Facter to use it.

branan:
This one's going to take a bit more review than the others :)

I'll start reading through it

MikaelSmith:
This is pulled from https://github.com/puppetlabs/facter/blob/master/lib/CMakeLists.txt#L8-L16; I found we didn't use the SHA1 result, just LIBFACTER_COMMIT.

branan:
Sweet, good to know this isn't whole-cloth :)

branan:
These are inconsistently named - `git_describe` `git_get_exact_tag` and `get_git_head_revision`. I know some of that comes from git, but would be nice to normalize our interface to the git info we want

MikaelSmith:
Those all come from a 3rd party script we've adopted. We could update them, but I'm not sure it's worth the inconsistency with the original.

PR #32, opened 2015-06-23
(maint) Update appveyor to Boost 1.57

MikaelSmith:
Updates appveyor to use Boost 1.57, and mirror the Facter config.


branan:
This looks reasonable. Should the CMake version bump PR be merged before we hit the button on this one?

MikaelSmith:
It doesn't need to be. The requirement is more documentation than functional with respect to Boost, because we're already using a newer cmake version in CI.

PR #35, opened 2015-06-24
(maint) Add pthread dependency for boost.thread

MikaelSmith:
Some systems (at least ArchLinux) require pthread to be linked for
Boost.Thread. Add that dependency on Unix systems so other projects will
pick it up.


branan:
CMake actually has a "FindThreads" that will pick up the system threading library/flags no matter what platform you're on. Will make our life easier/saner on weird unices.

http://www.cmake.org/cmake/help/v3.2/module/FindThreads.html

PR #36, opened 2015-06-25
(docs) Add README for json_container

parisiale:
[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ globally. This can be used as follows:
 
 ### Using Catch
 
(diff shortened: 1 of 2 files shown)


MikaelSmith:
Should we add a link from the top-level README? This seems like a good pattern to start.

Also, can use [skip ci] to avoid the Travis blockage.

parisiale:
Thanks, I didn't know about [skip ci].

I've added the link to the new "Using JsonContainer" section.
We should probably add docs to all components and have an index on the top of the main README.

PR #38, opened 2015-06-29
(maint) Improved tests for file and directory utility functions.

Magisus:


file_util/tests/directory_utils_test.cc
```diff
@@ -25,7 +25,7 @@ namespace leatherman { namespace file_util {
             REQUIRE(file_contents.find("3\n") != file_contents.end());
         }
 
-	SECTION("can find a file to match a pattern") {
+	    SECTION("can find a file to match a pattern") {
```
MikaelSmith: This indent looks out of place.

MikaelSmith: It would be nice to make this check an actual regex, like `[0-1]`.

MikaelSmith:
:+1:

PR #39, opened 2015-06-29
(LTH-18) Extract string and time utility functions to Leatherman

Magisus:
Extracts and standardizes scattered string and time utility functions from cthun-agent, cthun-client, and pegasus.

Currently, there are no unit tests for these methods in any of the original projects. Do we want some for Leatherman?

### DIFF
--- a/util/CMakeLists.txt
+++ b/util/CMakeLists.txt
@@ -1,5 +1,14 @@
-find_package(Boost 1.54 REQUIRED)
+find_package(Boost 1.54 REQUIRED date_time chrono system)
 
+add_leatherman_deps(${Boost_LIBRARIES})
 add_leatherman_includes("${Boost_INCLUDE_DIRS}")
 
 add_leatherman_headers(inc/leatherman)
+
+if(WIN32)
+    set(PLATFORM_SRCS "src/windows/time.cc")
+else()
+    set(PLATFORM_SRCS "src/posix/time.cc")
+endif()
+
+add_leatherman_library(src/strings.cc src/time.cc ${PLATFORM_SRCS})
--- a/util/inc/leatherman/util/strings.hpp
+++ b/util/inc/leatherman/util/strings.hpp
@@ -7,6 +7,7 @@
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/algorithm/string/compare.hpp>
 #include <string>
+#include <vector>
 
 namespace leatherman { namespace util {
 
@@ -25,5 +26,21 @@ namespace leatherman { namespace util {
                 return boost::lexicographical_compare(s1, s2, boost::is_iless());
             }
         };
-}}  // namespace leatherman::util
 
+        /**
+         * @return Returns the "s" string in case of more than one thing,
+         *      an empty string otherwise.
+         */
+        std::string plural(int num_of_things);
+
+        /**
+         * @return Returns the "s" string if vector contains more than one item,
+         *      an empty string otherwise.
+         */
+        template<typename T>
+        std::string plural(std::vector<T> const& things);
+
+        /** @return Returns universally unique identifier string. */
+        std::string get_UUID();
+
+}}  // namespace leatherman::util
--- a/util/inc/leatherman/util/time.hpp
+++ b/util/inc/leatherman/util/time.hpp
@@ -1,0 +1,36 @@
+/**
+ * @file
+ * Declares utility functions for dealing with time.
+ */
+#pragma once
+
+#include <string>
+#include <ctime>
+
+namespace leatherman { namespace util {
+
+    /**
+     * Adds the specified expiry_minutes to the current time and returns
+     * the related date time string in UTC format.
+     * @return Returns an empty string in case it fails to allocate the buffer.
+     */
+    std::string get_expiry_datetime(int expiry_minutes = 1);
+
+    /**
+     * Gets the current time in ISO8601 format
+     * @param modifier_in_secords Offset from the current time in seconds
+     * @return Returns the current time plus the modifier
+     */
+    std::string get_ISO8601_time(unsigned int modifier_in_seconds = 0);
+
+    /** @return Returns the current datetime string in the %Y%m%d_%H%M%S format */
+    std::string get_date_time();
+
+    /**
+     * Turns a stored time into a local time with correction for timezones.
+     *  @param stored_time The time to be converted.
+     *  @param result The struct in which to store the local time.
+     */
+    void get_local_time(std::time_t* stored_time, std::tm* result);
+
+}}  // namespace leatherman::util
--- a/util/src/posix/time.cc
+++ b/util/src/posix/time.cc
@@ -1,0 +1,9 @@
+#include <leatherman/util/time.hpp>
+
+namespace leatherman { namespace util {
+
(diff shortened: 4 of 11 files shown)


MikaelSmith:
Windows isn't happy. You may be able to rewrite portions of that using Boost.Date_Time.

Magisus:
What's bizarre is it builds on my VM... O.o

I'll switch that out anyway, just actually surprised that it works there.

MikaelSmith:
That's interesting... I have no idea why.

Magisus:
So more ugliness: std::mutex is undefined by some versions of MinGW. Boost's equivalent should work exactly the same, but it emits a bunch of warnings of the form "dereferencing type-punned pointer will break strict-aliasing rules", which are reported as errors, and therefore prevent compilation.

Is there another cross-platform alternative, and/or a way to get the boost stuff working?

util/src/time.cc
```diff
@@ -0,0 +1,66 @@
...
+        // Get local time structure
+        get_local_time(&expiry_time, &expiry_time_info);
+
+        // Return the formatted string
+        if (strftime(expiry_time_buffer, 80, "%FT%TZ", &expiry_time_info) == 0) {
+            // invalid buffer
+            return "";
+        }
+
+        return std::string { expiry_time_buffer };
```
MikaelSmith: There's a trick I've used in similar places to avoid doing the string copy. You can do
```
string buffer(80, '\0');
strftime(&buffer[0], ...)
```
This is guaranteed to be safe by the C++11 standard.

util/CMakeLists.txt
```diff
@@ -1,5 +1,14 @@
-find_package(Boost 1.54 REQUIRED)
+find_package(Boost 1.54 REQUIRED system thread chrono date_time)
```
MikaelSmith: For static libraries, this order needs to be: `x depends_on y depends_on z`. So this should probably be `date_time chrono system`. I think `thread` is no longer needed.

util/inc/leatherman/util/time.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Erases the current line on stdout, places the cursor at the
+     * beginning of it, and displays a progress message.
+     */
+    void displayProgress(double percent, int len,
+                         std::string status = "in progress");
+
+    /** @return Return the current time in ISO8601 format */
+    std::string getISO8601Time(unsigned int modifier_in_seconds = 0);
```
MikaelSmith: What does `modifier_in_seconds` do?

util/src/time.cc
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    void displayProgress(double percent, int len, std::string status) {
+        printf("%c[2K", 27);
```
MikaelSmith: This adds colorization, we should make this work on Windows or note it won't be supported. @peterhuene I'm not particularly familiar with terminal colorization, how do I read this?

@Magisus you can look at the colorization handling in logging for some guidance.

MikaelSmith: Lost this comment in the force push, but it's still valid.

Magisus: So, I was able to make this work cross platform by not actually clearing the line, but just printing over it. The way this is implemented, the length seems to be constant, or else only increasing in length (going from 99% to 100%). Can anyone think of any time when this would not be the case?

There is not really a clean way to do it on Windows (messiness involving manually specifying the cursor position or printing a backspace a certain number of times), but we can try one of those if we think the line really needs to be cleared.

As a side note, is this even being used currently anywhere? I couldn't find any references in any of the projects...

util/inc/leatherman/util/strings.hpp
```diff
@@ -25,5 +26,21 @@ namespace leatherman { namespace util {
...
+         *      an empty string otherwise.
+         */
+        std::string plural(int num_of_things);
+
+        /**
+         * @return Returns the "s" string if vector contains more than one item,
+         *      an empty string otherwise.
+         */
+        template<typename T>
+        std::string plural(std::vector<T> things);
```
peterhuene: When we port, we need to fix these extraneous copies of arguments.

util/src/posix/time.cc
```diff
@@ -0,0 +1,14 @@
+#include <leatherman/util/time.hpp>
+
+namespace leatherman { namespace util {
+
+    void get_local_time(time_t* stored_time, tm* result){
+        localtime_r(stored_time, result);
+    }
+
+    void clear_current_line(){
+        printf("%c[2K", 27);
```
peterhuene: `printf` in C++?  This will not interact well with a buffered iostream.

util/src/strings.cc
```diff
@@ -0,0 +1,23 @@
...
+#include <boost/uuid/uuid_generators.hpp>
+
+namespace leatherman { namespace util {
+
+    std::string plural(int num_of_things) {
+        return num_of_things > 1 ? "s" : "";
+    }
+
+    template<>
+    std::string plural<std::string>(std::vector<std::string> things) {
```
MikaelSmith: Oh yeah, should be a const&.

util/src/strings.cc
```diff
@@ -0,0 +1,23 @@
...
+        return plural(things.size());
```
peterhuene: We're making a copy of a vector to take the size?!

util/src/posix/time.cc
Magisus: If we do away with erasing the line, we get both cross-platform support and no printf's in one go (see my other comment on one of the above outdated diff).

util/inc/leatherman/util/time.hpp
```diff
@@ -0,0 +1,48 @@
...
+     * @return Returns an empty string in case it fails to allocate the buffer.
+     */
+    std::string getExpiryDatetime(int expiry_minutes = 1);
+
+    /**
+     * Erases the current line on stdout, places the cursor at the
+     * beginning of it, and displays a progress message.
+     */
+    void displayProgress(double percent, int len,
+                         std::string status = "in progress");
```
MikaelSmith: This function seems to be unused; let's just get rid of it.

util/src/time.cc
```diff
@@ -0,0 +1,65 @@
...
+        if (strftime(&expiry_time_buffer[0], 80, "%FT%TZ", &expiry_time_info) == 0) {
+            // invalid buffer
+            return "";
+        }
+
+        expiry_time_buffer.resize(strlen(&expiry_time_buffer[0]));
+        return expiry_time_buffer;
+    }
+
+    void displayProgress(double percent, int len, std::string status) {
```
MikaelSmith: @parisiale you added displayProgress in another project without using it. Was this something you expected to use in the future, or an accident?

MikaelSmith:
:+1:

util/inc/leatherman/util/time.hpp
```diff
@@ -0,0 +1,41 @@
+/**
+ * @file
+ * Declares utility functions for dealing with time.
+ */
+#pragma once
+
+#include <string>
+#include <time.h>
```
MikaelSmith: `time.h` is deprecated in C++ (http://www.cplusplus.com/reference/clibrary/), we should use `ctime`, and prefix its contents with `std::`.

MikaelSmith:
:+1:

PR #41, opened 2015-07-01
(LTH-14) Extract facter's curl wrapper to Leatherman

Magisus:
Extracts facter's http/client, http/request, and http/response classes to Leatherman. These provide a wrapper for creating and handling curl requests.

@nicklewis So this is what the curl wrapper classes currently look like. Do these seem applicable to your use of curl in the deployer-client? They can be extended with further functions if need be.


curl/src/client.cc
```diff
@@ -0,0 +1,388 @@
...
+
+        // If the header starts with "HTTP/", then we have the response status
+        if (input.starts_with("HTTP/")) {
+            // Reset the response buffer
+            ctx->response_buffer.clear();
+
+            // Parse out the error code
+            static boost::regex regex("HTTP/\\d\\.\\d (\\d\\d\\d).*");
+            int status_code = 0;
+            if (file_util::re_search(input.to_string(), regex, &status_code)) {
```
nicklewis: I think this should be `util`, not `file_util`?

nicklewis:
I hacked up a branch that uses this and yep, it fits our use case pretty well. I'll file a ticket about the one missing feature (setting SSL context).

Magisus:
Appveyor is failing because we are currently not installing or pointing to curl. Do we want to update the script to account for this?

nicklewis:
[LTH-25](https://tickets.puppetlabs.com/browse/LTH-25) tracks adding support for specifying SSL certs.

Magisus:
From what I could find in deployer client, I added some code that should support CA certs. @nicklewis  Are you using the client cert+key thing anywhere yet? If so, can you point out to me where that is? If not, is there another example of the functionality you want that you could point me to?

nicklewis:
We're not currently using client certs. I think it should look basically the same as the CA cert, except the settings are `CURLOPT_SSLCERT` and `CURLOPT_SSLKEY`. They both need to be specified together.

http://curl.haxx.se/libcurl/c/simplessl.html

That example seems to do some more work to support different cert types, etc, that we probably don't need.

curl/inc/leatherman/curl/request.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+        /**
+         * Sets the timeout for connecting to the remote host, in milliseconds.
+         * @param value The timeout value, in milliseconds.
+         */
+        void connection_timeout(long value);
+
+        /**
+         * Sets the CA certificate for HTTPS requests.
+         * @param ca_cert A string representing the certificate.
```
nicklewis: This seems to imply the string contains the certificate content; it's actually the file path.

curl/inc/leatherman/curl/client.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+     private:
+        request _req;
+    };
+
+    /**
+     * Implements a client for HTTP.
+     * Note: this class is not thread-safe.
+     */
+    struct client
```
nicklewis: Should this be CamelCase a la #42?

Magisus: Heh, good question. I feel like things are still undecided on that front. But yeah I'll update it for now.

curl/src/request.cc
```diff
@@ -0,0 +1,138 @@
...
+    {
+        return _connection_timeout;
+    }
+
+    void request::connection_timeout(long value)
+    {
+        _connection_timeout = value < 0 ? 0 : value;
+    }
+
+    void request::ca_info(std::string caCert)
```
nicklewis: This seems like the sort of thing that should actually be set on the client rather than the request, since it's probably going to be the same for every request. I would imagine that if you're talking to multiple services with different certs, those are two separate clients.

appveyor.yml
```diff
@@ -10,9 +10,12 @@ install:
   - ps: wget 'https://s3.amazonaws.com/kylo-pl-bucket/boost_1_57_0-x86_64_mingw-w64_4.8.3_win32_seh.7z' -OutFile "$pwd\boost.7z"
   - ps: 7z.exe x boost.7z -oC:\tools | FIND /V "ing  "
 
+  - ps: wget 'https://s3.amazonaws.com/kylo-pl-bucket/curl-7.42.1-x86_64_mingw-w64_4.8.3_win32_seh.7z' -OutFile "$pwd\curl-7.42.1-x86_64_mingw-w64_4.8.4_win32_seh.7z"
+  - ps: 7z.exe x "curl-7.42.1-x86_64_mingw-w64_4.8.4_win32_seh.7z" -oC:\tools | FIND /V "ing "
+
 build_script:
   - ps: mv "C:\Program Files (x86)\Git\bin\sh.exe" "C:\Program Files (x86)\Git\bin\shxx.exe"
-  - ps: cmake -G "MinGW Makefiles" -DBOOST_ROOT="C:\tools\boost_1_57_0-x86_64_mingw-w64_4.8.3_win32_seh" -DBOOST_STATIC=ON -DBOOST_NOWIDE_SKIP_TESTS=ON .
+  - ps: cmake -G "MinGW Makefiles" -DBOOST_ROOT="C:\tools\boost_1_57_0-x86_64_mingw-w64_4.8.3_win32_seh" -DBOOST_STATIC=ON -DBOOST_NOWIDE_SKIP_TESTS=ON -DCMAKE_PREFIX_PATH="C:\tools\curl-7.42.1-x86_64_mingw-w64_4.8.4_win32_seh" .
```
MikaelSmith: Ah, this won't be 4.8.4, it'll be 4.8.3. The directory name matches the server's file name.

MikaelSmith:
You ended up rolling the class renaming into the LTH-25 commit; makes it a little harder to pick apart.

MikaelSmith:
Looks like the same for getting appveyor working; could you split the LTH-25 work out?

Magisus:
So how many commits do you want total? Those bits in the first, and then the LTH-25 stuff specifically in its own? Or separate ones for those two tasks?

MikaelSmith:
Either way is fine. Probably all bits in the 1st except LTH-25 makes sense.

MikaelSmith:
I'd like to add tests, but in this case that might be outside the scope of the ticket. Let's file a new one to extend testing on the curl wrapper.

MikaelSmith:
:+1:

MikaelSmith:
Needs a rebase.

curl/inc/leatherman/curl/client.hpp
MikaelSmith: @peterhuene and I ended up badgering people into snake_case, so we'll stick with it for now.

MikaelSmith:
@nicklewis do the SSL cert extensions seem to fit your needs?

nicklewis:
I think it would make sense for client cert+key to *also* be specified on the client rather than the request, but it doesn't matter much for our project specifically, since we only use a CA cert, no client certs.

Magisus:
It's a fairly straightforward change, if usage-wise you think it make more sense. For the CA cert I just provided a setter; is that an acceptable pattern for the API on these guys? Or is there an approach you guys would find more elegant/usable?

nicklewis:
I think a setter is fine.

curl/inc/leatherman/curl/client.hpp
```diff
@@ -0,0 +1,217 @@
...
+         * @param req The HTTP request to perform.
+         * @return Returns the HTTP response.
+         */
+        response put(request const& req);
+
+        /**
+         * Sets the path to the CA certificate file.
+         * @param cert_file The path to the CA certificate file.
+         */
+        void set_ca_cert(std::string cert_file);
```
nicklewis: The names of these methods are a little confusing. For CA, `set_ca_cert` is public and `set_ca_info` is the private method for the context, but for client certs it's reversed: `set_client_ssl_info` is public and `set_client_cert` is private. It would be clearer to be consistent about `_cert` vs. `_info`. I don't have a strong opinion on which ought to be which though.

curl/src/client.cc
```diff
@@ -0,0 +1,423 @@
...
+        LOG_DEBUG("request completed (status %1%).", res.status_code());
+
+        // Set the body of the response
+        res.body(move(ctx.response_buffer));
+        return res;
+    }
+
+    void client::set_ca_cert(std::string cert_file)
+    {
+        _ca_cert = cert_file;
```
peterhuene: Missing a `std::move`.

curl/src/client.cc
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    void client::set_client_cert(std::string client_cert, std::string client_key)
+    {
+        _client_cert = client_cert;
```
peterhuene: Missing `std::move`s.

curl/inc/leatherman/curl/client.hpp
peterhuene: huzzah!

curl/src/client.cc
```diff
@@ -0,0 +1,423 @@
...
+        res.body(move(ctx.response_buffer));
+        return res;
+    }
+
+    void client::set_ca_cert(std::string cert_file)
+    {
+        _ca_cert = cert_file;
+    }
+
+    void client::set_client_cert(std::string client_cert, std::string client_key)
```
peterhuene: Also, I don't like seeing `std::` namespace prefixes unless it's resolving an ambiguity.  It's verbose and generally things in `std` are considered an integral part of the C++ "experience".

curl/inc/leatherman/curl/client.hpp
```diff
@@ -0,0 +1,217 @@
...
+     * The exception for HTTP requests.
+     */
+    struct http_request_exception : http_exception
+    {
+        /**
+         * Constructs an http_request_exception.
+         * @param req The HTTP request that caused the exception.
+         * @param message The exception message.
+         */
+        http_request_exception(request req, std::string const &message) :
```
peterhuene: This is an existing issue.  Taking the message by `const&` means the `std::move` is ineffective.  We should take `message` by value.

curl/src/client.cc
pcarlisle: Really? I'm pretty new to this but that contradicts the overwhelming majority of advice I've read on that subject.

Magisus: I've seen both in our C++ projects. At the moment, I'm pretty sure I'm explicitly writing `std::` almost everywhere in Leatherman, but it's something I could think about changing if there is consensus on a preferred way to do it.

peterhuene: Either the parameter needs to be `const&` and we always copy (arguably better for assignment to keep existing buffer of sufficient size), or it needs to be by value and the caller decides to move or copy. As is now, there is a needless extra copy.

peterhuene: Erm, whoops, responded to the wrong comment. Silly phone. Let me respond when I get back to my computer.

peterhuene: To be clear: in headers one *always* qualifies the name because it is extremely bad form to introduce using statements in headers. In source files, a `using namespace std;` is not only acceptable, but I actively encourage it as it makes the source more readable, requires less typing, and generally your names should not be conflicting with anything in the std namespace. If there are ambiguities, qualifying the name then makes sense to resolve that ambiguity; otherwise it's needless.

curl/inc/leatherman/curl/client.hpp
Magisus: So, what is the best way to resolve this here? I assume your above comment was actually relating to this. 

Sorry, still new to all the move/copy stuff and best practices thereabouts.

peterhuene: Here `message` should be taken by value. For the CA members, we should probably take by `const&` and just do a copy (i.e. we don't need a `move` for those).

peterhuene: After talking with Maggie, I'm smoking crack here.  Ignore my suggestion about changing `message`.  The base type doesn't take the parameter by values, this is not what we should be doing.  I think I saw pass-by-const-ref, quickly glanced at the initializer list and saw the second item did a `std::move` and it invoked my wrong-o-meter without realizing that line was moving the request, not the message :/  Sorry about that!

With regards to the CA member functions, though, we should take those arguments by `const&` and not move or by-value and move.  I have no strong opinions on either as there are trade-offs for each.  Let's stick with taking them by `const&` and not moving for now.

curl/inc/leatherman/curl/client.hpp
peterhuene: This member should take the argument by `const&` since it isn't moving the argument in its implementation.

curl/inc/leatherman/curl/client.hpp
```diff
@@ -0,0 +1,217 @@
...
+         * @param cert_file The path to the CA certificate file.
+         */
+        void set_ca_cert(std::string cert_file);
+
+        /**
+         * Set client SSL certificate and key.
+         * @param client_cert The path to the client's certificate file.
+         * @param client_key The path to the client's key file.
+         */
+        void set_client_cert(std::string client_cert, std::string client_key);
```
peterhuene: This member should take the arguments by `const&` since it isn't moving either argument in its implementation.

MikaelSmith:
:+1:

PR #42, opened 2015-07-02
(maint) Renamed types to use CamelCase throughout

Magisus:
I changed all the ones I could find, if any one sees any that I missed I'll add them. 

Once we merge this, most of the projects that use Leatherman will need updating to reflect the rename.

### DIFF
--- a/file_util/inc/leatherman/file_util/file.hpp
+++ b/file_util/inc/leatherman/file_util/file.hpp
@@ -68,17 +68,17 @@ namespace leatherman { namespace file_util {
      */
     std::string shell_quote(std::string path);
 
-    struct file_copy {
+    struct FileCopy {
         boost::filesystem::path source;
         std::string relativeName;
     };
 
-    using file_list = std::vector<file_copy>;
+    using FileList = std::vector<FileCopy>;
 
     /**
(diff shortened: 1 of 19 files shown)


MikaelSmith:
Definitely brings a contrast to the stl and boost names.

Magisus:
There's still time to go the other way... :P

peterhuene:
I'm not a fan :/ but ¯\_(ツ)_/¯

MikaelSmith:
I'm finding I have a stronger opinion about it than I thought.

Magisus:
Michael's argument about STL and Boost is pretty strong within the context of Leatherman specifically, since it is designed to be used a lot like those libraries, as a "3rd party" set of utilities. Whether or not the various teams change things in their respective projects, Leatherman needs to be internally consistent. Which means no matter what we do, someone is going to cringe whenever they use it, and it's going to look weird with their code. 

So the question is, who gets to cringe and who gets to enjoy seamless meshing? In terms of the projects that want to integrate Leatherman, is there one where we care about consistency more for any reason?

ploubser:
Since there are such strong opinions about this, let's just go with the suggested STL/Boost style.

Our team will make the changes to our projects over time.

PR #46, opened 2015-07-07
(maint) Use TravisCI container-based builds

MikaelSmith:
Enable TravisCI container-based builds by avoiding (and explicitly
disabling) sudo and using the APT addon instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,45 +1,55 @@
 language: cpp
 compiler:
   - gcc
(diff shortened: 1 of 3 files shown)


kylog:
Cool! I didn't know about 'addons' for Travis CI.

MikaelSmith:
I think I also need to use 'pip install --user' to avoid needing sudo there.

PR #50, opened 2015-07-09
(LTH-13) Extract facter's execution functions to Leatherman

Magisus:
Extracts facter's execution class and various associated utilities to Leatherman.

This is my first stab at extricating these classes. Quite a few utility things had to come with. Do we want to extract these entirely as well, or leave copies in facter for use in unrelated classes there?


Magisus:
Ideas for why Travis cppcheck is failing, but the same code doesn't fail in facter or on my machine?

MikaelSmith:
The Travis cppcheck job performs static analysis via `make check`. Right now it's using cppcheck-1.69, which has a problem with that piece of code. I'm probably going to put up a cppcheck-1.68 we can use and file a bug against cppcheck.

MikaelSmith:
Needs a rebase.

util/inc/leatherman/util/environment.hpp
```diff
@@ -0,0 +1,69 @@
...
+         * Gets the platform-specific path separator.
+         * @return Returns the platform-specific path separator.
+         */
+        static char get_path_separator();
+
+        /**
+         * Gets the platform-specific search program paths.
+         * @return Returns the platform-specific program search paths.
+         */
+        static std::vector<std::string> const& search_paths();
```
MikaelSmith: I'd rather not move the search_paths to Leatherman, those are unique to Facter (for now).

MikaelSmith: Actually, these are somewhat Facter-agnostic. @branan @peterhuene thoughts on this? We essentially just grab PATH, and on Linux add `/sbin` and `/usr/sbin`.

MikaelSmith:
Facter has environment unit tests we should pull in.

MikaelSmith:
Same for execution. We should pull in those tests and related fixtures.

util/inc/leatherman/util/environment.hpp
Magisus: So far as I can see, the method is only used to provide default arguments to a couple of the functions (which, expand_command). If we remove this, we'll need to supply something else there.

peterhuene: It doesn't "hurt" per se, but I've always thought it a little weird that facter checks these locations by default rather than always relying on PATH.

execution/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+add_leatherman_headers(inc/leatherman)
+if(WIN32)
+    add_leatherman_library(src/execution.cc src/windows/execution.cc)
+else()
+    add_leatherman_library(src/execution.cc src/posix/execution.cc)
+endif()
+
+if(WIN32)
+    set(PLATFORM_TESTS tests/windows/execution.cc)
```
MikaelSmith: Facter has a windows/environment.cc test as well at https://github.com/puppetlabs/facter/blob/master/lib/tests/util/windows/environment.cc.

execution/tests/log_capture.hpp
```diff
@@ -0,0 +1,35 @@
+#pragma once
```
MikaelSmith: This utility is used in several tests that will be split between different modules in Leatherman. I wonder if there's a model we could use for sharing test utilities. @branan have any thoughts?

util/src/strings.cc
```diff
@@ -3,21 +3,39 @@
...
     }
 
-    std::string get_UUID() {
+    string get_UUID() {
         static boost::uuids::random_generator gen;
         boost::uuids::uuid uuid = gen();
         return boost::uuids::to_string(uuid);
     }
 
+    void each_line(string const& s, function<bool(string&)> callback) {
```
MikaelSmith: New functions should get new tests. Also, the formatting looks odd.

MikaelSmith:
Ready for review?

Magisus:
Yes, though there's still one test failing when I merge back into facter. Not sure if the issue is in this code, or in the way I merged it...

.gitignore
```diff
@@ -12,3 +12,4 @@ compile_commands.json
 /.idea/
 /*build/
 /dynamic_library/tests/fixtures.hpp
+/execution/tests/fixtures.hpp
```
MikaelSmith: You could probably make this a wildcard, like `/*/tests/fixtures.hpp`.

execution/inc/leatherman/execution/execution.hpp
```diff
@@ -0,0 +1,311 @@
...
+     * @param directories The directories to search.
+     * @return Returns the full path or empty if the file could not be found.
+     */
+    std::string which(std::string const& file, std::vector<std::string> const& directories = lth_util::environment::search_paths());
+
+    /**
+     * Expands the executable in the command to the full path.
+     * @param command The command to expand.
+     * @param directories The directories to search.
+     * @return Returns the expanded command if the executable was found or the original command if not.
```
MikaelSmith: This comment changed in https://github.com/puppetlabs/facter/commit/104a9371b98e71aae11965dacff344c26532f8a3.

execution/inc/leatherman/execution/option_set.hpp
```diff
@@ -0,0 +1,228 @@
...
+ * @file
+ * Declares utility type for passing a set of options.
+ */
+#pragma once
+
+#include <numeric>
+#include <string>
+#include <initializer_list>
+
+namespace leatherman { namespace execution {
```
MikaelSmith: option_set was previously in the util namespace, as it is a general-purpose utility. I'd probably keep it there.

MikaelSmith:
For now add fixes as new commits, so I can keep track of changes. This is a pretty large PR. We can squash them later when ready to merge.

execution/tests/environment.cc
```diff
@@ -0,0 +1,90 @@
+#include <catch.hpp>
```
MikaelSmith: Why is this file in execution, rather than util?

execution/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+add_leatherman_headers(inc/leatherman)
+if(WIN32)
+    add_leatherman_library(src/execution.cc src/windows/execution.cc)
+else()
+    add_leatherman_library(src/execution.cc src/posix/execution.cc)
+endif()
+
+if(WIN32)
+    set(PLATFORM_TESTS tests/windows/execution.cc tests/windows/environment.cc)
```
MikaelSmith: All the environment tests should be in util, not execution.

execution/tests/environment.cc
Magisus: That's a good question... xD

util/inc/leatherman/util/scoped_descriptor.hpp
```diff
@@ -0,0 +1,27 @@
+/**
```
MikaelSmith: This only works on POSIX. I'm not too tied to putting it in a posix subdirectory, but @peterhuene may have thoughts on that.

peterhuene: The subdirectories were a way to simply make it clear that this particular type was specific to certain platforms (it was also in a `posix` namespace if memory serves).  It's simply to make it clear that this is not something you can include on every platform.  I'm not particular married to this convention, but I found it to be useful, personally.

Magisus:
Since there is now a test for one method in strings.cc, coveralls is complaining about the rest that don't have tests. Do we care enough to write tests for those (plural(), getUUID())?

MikaelSmith:
I was going to write simple tests later if you didn't, so let's go ahead and add some.

Michael

Magisus:
Tests written, can be found at https://github.com/puppetlabs/leatherman/pull/54.

MikaelSmith:
Needs to be rebased now.

MikaelSmith:
Looks like you need to add the environments tests to a test in util/CMakeLists.txt.

util/inc/leatherman/util/scoped_descriptor.hpp
MikaelSmith: @Magisus I think I agree that this is clearer in util/posix. We may want to re-arrange posix-specific and windows-specific utilities sometime to be more consistent (we have a windows module and then a posix-specific class in the util module).

Magisus: Do we want to change the namespace too, or just add the subdirectory?

MikaelSmith: Yeah, let's namespace for now. Makes it more obvious to anyone using it that it will just fail on Windows.

MikaelSmith:
:+1: squash commits and I'll merge it.

PR #51, opened 2015-07-13
(LTH-28) fix build on FreeBSD

mmoll:


dynamic_library/CMakeLists.txt
```diff
@@ -3,9 +3,12 @@ leatherman_dependency(util)
 if(WIN32)
     leatherman_dependency(windows)
 else()
-  if(NOT CMAKE_SYSTEM_NAME MATCHES "OpenBSD")
+  if(NOT CMAKE_SYSTEM_NAME MATCHES "FreeBSD|OpenBSD")
       add_leatherman_deps(dl)
   endif()
+  if(CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
```
branan: This logic should probably be in `cmake/cflags.cmake`? My understanding is that FreeBSD uses GCC or Clang depending on the OS version, and we have support for those compiler explicitly. If something about FreeBSD is causing those flags (including -fPIC) to not get picked up it sould be addressed there.

branan: OK, it looks like we only add -fPIC with GCC, not with Clang. You can probably just move this entire "if FreeBSD" block into the Clang block in that file.

MikaelSmith:
Also I've put up a PR for [contributor guidelines](https://github.com/puppetlabs/leatherman/pull/52), please take a look.

mmoll:
I think the comments are addressed now. My JIRA username is mmoll and it seems I can't open issues in the LTH project...

MikaelSmith:
Working on getting the JIRA issue fixed up.

MikaelSmith:
Ok, I think JIRA should be setup correctly now.

mmoll:
thanks, done!

PR #54, opened 2015-07-14
(maint) Added tests for string utilities.

Magisus:


util/src/strings.cc
```diff
@@ -6,7 +6,7 @@
 namespace leatherman { namespace util {
 
     std::string plural(int num_of_things) {
-        return num_of_things > 1 ? "s" : "";
+        return num_of_things == 1 ? "" : "s";
```
Magisus: If we're going to bother to pluralize, we might as well be grammatically correct: "0 things", not "0 thing".

MikaelSmith: :+1:

util/tests/strings-test.cc
```diff
@@ -0,0 +1,32 @@
...
+        REQUIRE(plural(things) == "s");
+    }
+}
+
+TEST_CASE("strings::get_UUID", "[strings]") {
+    SECTION("returns a unique value each time") {
+        set<string> ids;
+        ids.insert(get_UUID());
+        for(int i = 0; i < 100; i++) {
+            string new_id = get_UUID();
```
MikaelSmith: Well, this test will pass 99.99999% of the time; good enough.

PR #55, opened 2015-07-15
(LTH-15) Extract facter's Ruby VM hosting to Leatherman

Magisus:
Moves facter's Ruby API to Leatherman.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,7 @@ install:
 
   - choco install -y mingw-w64 -Version 4.8.3 -source https://www.myget.org/F/puppetlabs
   - choco install -y cmake -Version 3.2.3
-  - SET PATH=C:\tools\mingw64\bin;%PATH%
+  - SET PATH=C:\Ruby21-x64\bin;C:\tools\mingw64\bin;%PATH%
 
   - ps: wget 'https://s3.amazonaws.com/kylo-pl-bucket/boost_1_57_0-x86_64_mingw-w64_4.8.3_win32_seh.7z' -OutFile "$pwd\boost.7z"
   - ps: 7z.exe x boost.7z -oC:\tools | FIND /V "ing  "
@@ -15,7 +15,7 @@ install:
 
 build_script:
   - ps: mv "C:\Program Files (x86)\Git\bin\sh.exe" "C:\Program Files (x86)\Git\bin\shxx.exe"
-  - ps: cmake -G "MinGW Makefiles" -DBOOST_ROOT="C:\tools\boost_1_57_0-x86_64_mingw-w64_4.8.3_win32_seh" -DBOOST_STATIC=ON -DBOOST_NOWIDE_SKIP_TESTS=ON -DCMAKE_PREFIX_PATH="C:\tools\curl-7.42.1-x86_64_mingw-w64_4.8.3_win32_seh" -DCURL_STATIC=ON .
+  - ps: cmake -G "MinGW Makefiles" -DBOOST_ROOT="C:\tools\boost_1_57_0-x86_64_mingw-w64_4.8.3_win32_seh" -DBOOST_STATIC=ON -DBOOST_NOWIDE_SKIP_TESTS=ON -DCMAKE_PREFIX_PATH="C:\tools\curl-7.42.1-x86_64_mingw-w64_4.8.3_win32_seh" -DCURL_STATIC=ON -DRUBY_LIBRARY="C:\Ruby21-x64\lib" .
   - ps: mv "C:\Program Files (x86)\Git\bin\shxx.exe" "C:\Program Files (x86)\Git\bin\sh.exe"
   - ps: mingw32-make
 
--- a/ruby/inc/leatherman/ruby/api.hpp
+++ b/ruby/inc/leatherman/ruby/api.hpp
@@ -1,0 +1,675 @@
+/**
+ * @file
+ * Declares the API imported from Ruby.
+ */
+#pragma once
+
+#include <cstdint>
+#include <string>
+#include <vector>
+#include <set>
+#include <functional>
+#include <memory>
+#include <initializer_list>
+
+#include <leatherman/dynamic_library/dynamic_library.hpp>
+
+namespace leatherman {  namespace ruby {
+
+    /*
+     * Parts of the MRI (Matz's Ruby Interpreter; a.k.a. CRuby) we use is documented here:
+     * https://github.com/ruby/ruby/blob/trunk/README.EXT
+     *
+     * Otherwise, the canonical documentation is unfortunately the MRI source code itself.
+     * A useful index of the various MRI versions can be found here:
+     * http://rxr.whitequark.org/mri/source
+     *
+     */
+
+    /**
+     * Represents a MRI VALUE (a Ruby object).
+     * VALUEs can be constants denoting things like true, false, or nil.
+     * They can also be encoded numerical values (Fixnum, for example).
+     * They can also be pointers to a heap-allocated Ruby object (class, module, etc).
+     * The Ruby garbage collector scans the main thread's stack for VALUEs to mark during garbage collection.
+     * Therefore, you may encounter "volatile" VALUES. These are marked simply to ensure the compiler
+     * does not do any optimizations that may prevent the garbage collector from finding them.
+     * This is likely not needed, but it isn't hurting us to do.
+     */
+    typedef uintptr_t VALUE;
+    /**
+     * See MRI documentation.
+     */
+    typedef intptr_t SIGNED_VALUE;
+    /**
+     * See MRI documentation.
+     */
+    typedef uintptr_t ID;
+
+    /**
+     * See MRI documentation. This is a complex struct type; we only use it as an opaque object.
+     */
+    typedef void * rb_encoding_p;
+
+    /**
+     * Macro to cast function pointers to a Ruby method.
+     */
+    #define RUBY_METHOD_FUNC(x) reinterpret_cast<VALUE(*)(...)>(x)
+
+    /**
+     * Exception thrown when ruby library could not be loaded.
+     */
+    struct library_not_loaded_exception : std::runtime_error {
+        /**
+         * Constructs a library_not_loaded_exception.
+         * @param message The exception message.
+         */
+        explicit library_not_loaded_exception(std::string const& message);
+    };
+
+    /**
+     * Contains utility functions and the pointers to the Ruby API.
+     */
+    struct api
+    {
+        /**
+         * Destructs the Ruby API.
+         */
+        ~api();
+
+        /**
+         * Prevents the API from being copied.
+         */
+        api(api const&) = delete;
+        /**
+         * Prevents the API from being copied.
+         * @returns Returns this API.
+         */
+        api& operator=(api const&) = delete;
+        /**
+         * Prevents the API from being moved.
+         */
+        api(api&&) = delete;
+        /**
+         * Prevents the API from being moved.
+         * @return Returns this API.
+         */
+        api& operator=(api&&) = delete;
+
+        /**
+         * Gets the Ruby API instance.
+         * @return Returns the Ruby API instance or nullptr if the Ruby API is unavailable.
+         */
+        static api* instance();
+
+        /**
+         * Called to initialize the API.
+         * This should be done at the same stack frame where code is loaded into the Ruby VM.
+         */
+        void initialize();
+
+        /**
+         * Gets whether or not the API has been initialized.
+         * @return Returns true if the API has been initialized or false if it has not been initialized.
+         */
+        bool initialized() const;
+
+        /**
+         * Gets whether or not exception stack traces are included when formatting exception messages.
+         * @return Returns true if stack traces will be included in exception messages or false if they will not be.
+         */
+        bool include_stack_trace() const;
+
+        /**
+         * Sets whether or not exception stack traces are included when formatting exception messages.
+         * @param value True if stack traces will be included in exception messages or false if they will not be.
+         */
+        void include_stack_trace(bool value);
+
+        /**
+         * See MRI documentation.
+         */
+        ID (* const rb_intern)(char const*);
+        /**
+         * See MRI documentation.
+         */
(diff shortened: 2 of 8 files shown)


ruby/inc/leatherman/ruby/api.hpp
```diff
@@ -0,0 +1,653 @@
...
+        VALUE false_value() const;
+
+        // TODO: This is the only method that seems to be directly related to facter.
+        // TODO: What do we want to do with it?
+//        /**
+//         * Converts the given value to a corresponding Ruby object.
+//         * @param val The value to convert.
+//         * @return Returns a Ruby object for the value.
+//         */
+//        VALUE to_ruby(facter::facts::value const* val) const;
```
Magisus: This method seems facter-specific. How do we want to handle it?

ruby/src/api.cc
```diff
@@ -0,0 +1,509 @@
...
+#include <boost/filesystem/path.hpp>
+#include <boost/filesystem/operations.hpp>
+#include <sstream>
+
+using namespace std;
+using namespace leatherman::util;
+using namespace leatherman::execution;
+using namespace boost::filesystem;
+
+namespace lth_lib = leatherman::dynamic_library;
```
Magisus: If I try to use a `using namespace leatherman::dynamic_library` here, it complains every time `dynamic_library` appears as a type, with this: `error: must use 'struct' tag to refer to type 'dynamic_library' in this scope`. It's as if it's getting confused and thinks I'm referring to the namespace in places I shouldn't by. Why is this? I assume it has something to do with the struct and the namespace having the same name... is this inadvisable?

MikaelSmith:
This would be an interesting one to add tests to. Something simple could be doing an eval of some ruby code. A more advanced test would register a new ruby method that the ruby eval then calls.

ruby/src/api.cc
MikaelSmith: Which compiler complains, clang? I don't think it should be ambiguous, but I'm likely wrong.

Magisus: Yes, clang. I didn't try it with anything else, but I can. I'll let you know.

Magisus: gcc does it too. `error: ‘dynamic_library’ does not name a type`.

ruby/src/api.cc
```diff
@@ -0,0 +1,509 @@
...
+                // Clear RUBYOPT so that only our options are used.
+                environment::set("RUBYOPT", "");
+            }
+
+            ruby_options(sizeof(opts) / sizeof(opts[0]), const_cast<char**>(opts));
+        }
+
+        // Get the values for nil, true, and false
+        // We do this because these are not constant across ruby versions
+        _nil = rb_ivar_get(*rb_cObject, rb_intern("@facter_nil"));
```
Magisus: What doest `@facter_nil` refer to here? Is it something we can use, and if not what should be used instead?

MikaelSmith: I'm not really sure... @peterhuene?

MikaelSmith: That may be calling a value it assumes to be nil, in which case the name facter_nil could be changed to leatherman_nil.

peterhuene: That's exactly what it does.  The VALUE for nil changes between different releases of Ruby, so to make our code ruby version agnostic, we query for something that is known not to exist, which returns the correct VALUE for `nil`.  Changing it to something like `expected_to_be_nil` would be fine.

Magisus: I'll do that, makes it clearer what's happening.

ruby/inc/leatherman/ruby/api.hpp
MikaelSmith: This seems to make sense to move to ruby::module.

ruby/inc/leatherman/ruby/api.hpp
```diff
@@ -0,0 +1,664 @@
...
+        void unregister_data_object(VALUE obj) const
+        {
+            _data_objects.erase(obj);
+        }
+
+        /**
+         * Global flag to disable Ruby VM cleanup.
+         * This should be set to false in any forked child processes.
+         */
+        static bool cleanup;
```
MikaelSmith: I think this was removed, in https://github.com/puppetlabs/facter/commit/78be9ea963ec3fd24ecbf3ef2ba038c4b33b65b2#diff-fabbc6ab7f9f6a77c8cf164caa3860b6.

ruby/src/api.cc
```diff
@@ -0,0 +1,529 @@
...
+    {
+        static unique_ptr<api> instance = create();
+        return instance.get();
+    }
+
+    unique_ptr<api> api::create()
+    {
+        lth_lib::dynamic_library library = find_library();
+        if (!library.loaded()) {
+            LOG_WARNING("could not locate a ruby library");
```
MikaelSmith: These messages change; we can probably just modify Facter's message when it gets a nullptr to contain the same info.

ruby/src/api.cc
```diff
@@ -0,0 +1,529 @@
...
+    {
+        // First search for an already loaded Ruby.
+        auto library = find_loaded_library();
+        if (library.loaded()) {
+            return library;
+        }
+
+#ifdef LEATHERMAN_RUBY
+        // Ruby lib location was specified at compile-time, fix to that.
+        if (library.load(LEATHERMAN_RUBY)) {
```
MikaelSmith: This has some interesting repercussions on compiling Leatherman separately from Facter I need to consider. /cc @branan

ruby/src/api.cc
Magisus: So should facter handle the error in general (e.g. remove these logging calls from this class)? I edited the messages to no longer make references to facter functionality when I pulled this stuff out, but I could see that still being relevant info top-level, and we probably don't want to log anything twice...

MikaelSmith: We still need a method for handling info/debug messages (LTH-29), but for the warnings I think we should instead throw exceptions.

Magisus:
@MikaelSmith do you think we could talk about the approach for writing the tests for this at some point? I'm a little unclear on how to get started.

ruby/tests/api-test.cc
```diff
@@ -0,0 +1,132 @@
...
+    ruby->initialize();
+    REQUIRE(ruby->initialized());
+
+    SECTION("can detect class membership") {
+        REQUIRE(ruby->case_equals(ruby->eval("Integer"), ruby->eval("1")));
+        REQUIRE_FALSE(ruby->case_equals(ruby->eval("String"), ruby->eval("4")));
+    }
+}
+
+VALUE test_func(VALUE self) {
```
Magisus: Should this be somewhere else, or is it fine here?

MikaelSmith: Keeping function declarations local to where they're used works for me, until we have a reason not to. So this seems fine.

Magisus:
On my windows machine, Ruby can be found and loaded, and all the tests pass. What's up with AppVeyor? I am printing the location of what it believes is a valid ruby dll.

MikaelSmith:
Ah, it's always binary architecture. That Ruby is a 32-bit binary (Ruby 2.0 was the first build to support 64-bit on Windows). You can see installed Ruby versions at http://www.appveyor.com/docs/installed-software. For Facter we explicitly add Ruby 2.1 to the PATH (https://github.com/puppetlabs/facter/blob/master/appveyor.yml#L8).

ruby/inc/leatherman/ruby/api.hpp
```diff
@@ -0,0 +1,669 @@
...
+         */
+        VALUE false_value() const;
+
+        /**
+         * Looks up a constant based on the given names. The individual entries correspond to
+         * namespaces, as in {"A", "B", "C"} => A::B::C.
+         * @param names The names to lookup.
+         * @return Returns the value or raises a NameError.
+         */
+        VALUE lookup(std::initializer_list<std::string> const& names) const;
```
Magisus: This feels like something we could test, but I'm still a little unclear what it does/how to use it, despite the comment and some googling...

peterhuene: If you define module `Foo::Bar::Baz` in Ruby, you can use this member to find the VALUE representing the module: `lookup({ "Foo", "Bar", "Baz" })`.

MikaelSmith: Could you do something like `lookup({"RbConfig", "CONFIG['host_os']"})`?

peterhuene: No, that'd be `ruby.rb_hash_lookup(ruby.lookup({"RbConfig", "CONFIG"}), ruby.rb_str_new_cstr("host_os"));` (assuming we imported `rb_str_new_cstr`).

MikaelSmith:
I'll try to do a close review later this week. Also needs a rebase.

ruby/src/api.cc
MikaelSmith: If we want to make re-usable Leatherman binaries, this has to be configurable at run-time. That implies making it a configurable option on the api that others can specify at compile-time.

Magisus: So just turn it into a regular field of the class?

MikaelSmith: Making it an argument to the api::instance method doesn't seem like a great option - the whole point is that can be called repeatedly to get the same object, and we'd rather only specify Ruby loading once. I think this makes sense as a static variable that you set very early on, possibly static initialization (before entering the main() method), to configure the api class behavior.

MikaelSmith: So yeah, add a static member to the class that defaults to an empty string, and instead of `#ifdef` do if string is not empty.

appveyor.yml
```diff
@@ -15,7 +15,7 @@ install:
 
 build_script:
   - ps: mv "C:\Program Files (x86)\Git\bin\sh.exe" "C:\Program Files (x86)\Git\bin\shxx.exe"
-  - ps: cmake -G "MinGW Makefiles" -DBOOST_ROOT="C:\tools\boost_1_57_0-x86_64_mingw-w64_4.8.3_win32_seh" -DBOOST_STATIC=ON -DBOOST_NOWIDE_SKIP_TESTS=ON -DCMAKE_PREFIX_PATH="C:\tools\curl-7.42.1-x86_64_mingw-w64_4.8.3_win32_seh" -DCURL_STATIC=ON .
```
MikaelSmith: Why was CURL_STATIC dropped?

ruby/src/api.cc
```diff
@@ -0,0 +1,490 @@
...
+using namespace std;
+using namespace leatherman::util;
+using namespace leatherman::execution;
+using namespace boost::filesystem;
+
+namespace lth_lib = leatherman::dynamic_library;
+
+namespace leatherman { namespace ruby {
+
+        library_not_loaded_exception::library_not_loaded_exception(string const& message) :
```
MikaelSmith: Indentation appears off here.

ruby/tests/api-test.cc
```diff
@@ -0,0 +1,172 @@
...
+    SECTION("can correctly identify true and false values") {
+        REQUIRE(ruby->is_true(ruby->true_value()));
+        REQUIRE_FALSE(ruby->is_true(ruby->false_value()));
+        REQUIRE(ruby->is_false(ruby->false_value()));
+        REQUIRE_FALSE(ruby->is_false(ruby->true_value()));
+    }
+
+    SECTION("can correctly identify strings") {
+        REQUIRE(ruby->is_string(ruby->utf8_value("'I'm a string'")));
+        REQUIRE_FALSE(ruby->is_string(1));
```
MikaelSmith: This looks strange, I wouldn't expect `1` to be a valid `VALUE`. Maybe use nil/true/false, or eval to generate a number.

ruby/tests/api-test.cc
```diff
@@ -0,0 +1,172 @@
...
+        REQUIRE_FALSE(ruby->equals(ruby->true_value(), ruby->false_value()));
+    }
+
+    SECTION("can correctly test strings for equality") {
+        REQUIRE(ruby->equals(ruby->utf8_value("foo"), ruby->utf8_value("foo")));
+        REQUIRE_FALSE(ruby->equals(ruby->utf8_value("foo"), ruby->utf8_value("bar")));
+    }
+
+    SECTION("can correctly test numbers for equality") {
+        REQUIRE(ruby->equals(1, 1));
```
MikaelSmith: Again using integer literals. Are these expected to be safe comparisons?

Magisus: No, I don't think so. Sorry about this, I realized later on that the numbers weren't doing what I thought they were. Guess I forgot to go back and fix them...

ruby/src/api.cc
```diff
@@ -0,0 +1,490 @@
...
+    library_not_loaded_exception::library_not_loaded_exception(string const& message) :
+        runtime_error(message)
+    {
+    }
+
+#define LOAD_SYMBOL(x) x(reinterpret_cast<decltype(x)>(library.find_symbol(#x, true)))
+#define LOAD_ALIASED_SYMBOL(x, y) x(reinterpret_cast<decltype(x)>(library.find_symbol(#x, true, #y)))
+#define LOAD_OPTIONAL_SYMBOL(x) x(reinterpret_cast<decltype(x)>(library.find_symbol(#x)))
+
+    set<VALUE> api::_data_objects;
```
MikaelSmith: This is also a global, I'd group it and ruby_lib_location together.

MikaelSmith:
Other than the last comment, this looks good. I think you can squash it all up.

MikaelSmith:
Travis may be failing because of the outage at http://status.coveralls.io/.

ruby/inc/leatherman/ruby/api.hpp
```diff
@@ -0,0 +1,674 @@
...
+        /**
+         * Unregisters a data object.
+         * @param obj The data object to unregister.
+         */
+        void unregister_data_object(VALUE obj) const
+        {
+            _data_objects.erase(obj);
+        }
+
+        /** Specifies the location of the preferred Ruby library.
```
MikaelSmith: Could you put a newline after `/**` to follow other comment formatting?

PR #56, opened 2015-07-20
(LTH-27) Add unit tests for curl wrapper library

whopper:
This commit adds a suite of tests for Leatherman's
curl wrapper library. Tests utilize a new mock_curl struct
which mocks the libcurl API. We link against it instead of
real libcurl for testing purposes.


whopper:
As @MikaelSmith and I have discussed, I currently have some very shallow tests in place for requests, but they may not be super useful. I believe we're still on the fence on whether we'll keep them and add equivalent tests for responses, or remove them alltogether.

whopper:
Wat at coveralls, although `request.cc` coverage sure looks nice - maybe we *should* give `response.cc` the same treatment.

MikaelSmith:
I'd use that as a guide to add a few more tests: response's header support, and trying to test some failure cases.

whopper:
Also looks like we need to do something extra to fix up building on windows.

MikaelSmith:
That's a side-effect of using curl.h; I'd probably disable `-Werror=attributes` for that file using GCC diagnostic pragmas https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html.

MikaelSmith:
Linking errors now, I'm not surprised. You need to deal with dllexport/dllimport; you may be able to get away with defining CURL_STATICLIB for the curl test files. See https://github.com/bagder/curl/blob/master/include/curl/curl.h for where that define is used.

curl/CMakeLists.txt
```diff
@@ -13,4 +19,8 @@ endif()
 add_leatherman_library(src/client.cc src/request.cc src/response.cc)
 add_leatherman_headers(inc/leatherman)
 
-target_link_libraries(leatherman_curl "${CURL_LIBRARIES}")
+if(${WIN32})
+    add_definitions(-DCURL_STATICLIB)
```
MikaelSmith: We don't want to do this for everything, just tests. I lean towards defining it in the test files before including `curl/curl.h`.

curl/tests/mock_curl.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+#ifdef _WIN32
+#    pragma GCC diagnostic ignored "-Wattributes"
```
MikaelSmith: This line may no longer be needed because we're defining CURL_STATICLIB.

curl/tests/client_test.cc
```diff
@@ -0,0 +1,227 @@
+#include <catch.hpp>
+#include <leatherman/curl/client.hpp>
+#include <leatherman/curl/request.hpp>
+#include <leatherman/curl/response.hpp>
+#include "mock_curl.hpp"
```
MikaelSmith: AppVeyor's failing because `curl/curl.h` is included by the Leatherman headers. `mock_curl.hpp` should be included before them.

MikaelSmith:
If the tests are written, let's leave them until they become a problem.

curl/tests/mock_curl.hpp
```diff
@@ -0,0 +1,42 @@
...
+    http_method method = http_method::get;
+
+    curl_slist* header; // List of custom request headers to be passed to the server
+
+    std::string read_buffer; // Buffer to test reading the request body
+    std::string resp_body;   // Response body which should be written to a context using the write_body function callback
+};
+
+void set_error_mode(bool mode);
+static bool error_mode = false;
```
MikaelSmith: The error_mode flag should only exist in `mock_curl.cc`.

The only need we have for a global is when the curl_impl struct doesn't exist, i.e. during creation of the client. We use this to signal that an error should occur during client initialization.

I also lean towards making this a struct, so we can take advantage of RAII for automatic cleanup after the test.

curl/tests/client_test.cc
whopper: I'm afraid this isn't the only reason it's failing... still no luck.

MikaelSmith:
http://stackoverflow.com/a/4087557 looks like it covers the Travis build failure.

curl/tests/client_test.cc
```diff
@@ -0,0 +1,405 @@
...
+using namespace std;
+
+namespace leatherman { namespace curl {
+
+    struct mock_client : client {
+        curl_handle const& get_handle() { return client::get_handle(); }
+    };
+
+    TEST_CASE("curl::client") {
+        //cout << "\nNON ERROR TEST CLIENT INIT\n";
```
MikaelSmith: Remove

curl/tests/mock_curl.cc
```diff
@@ -0,0 +1,361 @@
...
+ * of the CURL API to ensure that our wrapper is making the calls
+ * to libcurl that we expect.
+ */
+
+/*
+ * Error struct implementations below... <for testing various libcurl failures>
+ * These are for the pre-client inits. The rest of the exceptions can use an
+ * enum inside of the curl_impl class
+ */
+error_mode test_failure_mode = success;
```
MikaelSmith: This should be declared `static`. It doesn't need to be exposed externally, that's what the methods are for.

curl/tests/client_test.cc
```diff
@@ -0,0 +1,405 @@
...
+            CURL* const& handle = test_client.get_handle();
+            auto test_impl = static_cast<curl_impl* const>(handle);
+            REQUIRE(test_impl->client_key == "key");
+        }
+    }
+
+    /* Exception handling tests below */
+
+    /*
+    TEST_CASE("curl::client fails to globally initialize libcurl") {
```
MikaelSmith: The global_init_error is tricky, as global init is done as part of static initialization in the curl helper. There's currently no way to force it to be re-initialized.

For now I'd replace this block with a comment about why we don't have a test covering it.

curl/tests/client_test.cc
```diff
@@ -0,0 +1,405 @@
...
+    */
+
+    TEST_CASE("curl::client exceptions") {
+        SECTION("client fails to initialize a libcurl easy session") {
+            curl_fail_init mock_error {easy_init_error};
+            REQUIRE_THROWS_AS(mock_client test_client, http_exception);
+        }
+
+        SECTION("client fails to perform a cURL request") {
+            mock_client test_client;
```
MikaelSmith: Everything except setting the test_failure_mode and the REQUIRE behavior can be pulled up into the TEST_CASE block if you move the 1st SECTION into its own TEST_CASE. That would make this a lot less verbose.

curl/tests/mock_curl.cc
```diff
@@ -0,0 +1,361 @@
...
+CURLcode curl_easy_setopt(CURL *handle, CURLoption option, ...)
+{
+    auto h = static_cast<curl_impl*>(handle);
+    va_list vl;
+    va_start(vl, option);
+
+    switch (option) {
+        case CURLOPT_HEADERFUNCTION:
+            // Set client::write_header as the function to be called as soon as mock curl has received header data.
+            if (h->test_failure_mode == curl_impl::error_mode::header_function_error) {
```
MikaelSmith: I don't see anything in the [curl documentation](http://curl.haxx.se/libcurl/c/CURLOPT_HEADERFUNCTION.html) saying this will return an error.

curl/tests/mock_curl.cc
```diff
@@ -0,0 +1,361 @@
...
+                va_end(vl);
+                return CURLE_COULDNT_CONNECT;
+            }
+            h->read_data = va_arg(vl, void*);
+            break;
+        case CURLOPT_URL:
+            // Set the mock curl URL as the URL specified in the request.
+            if (h->test_failure_mode == curl_impl::error_mode::set_url_error) {
+                va_end(vl);
+                return CURLE_COULDNT_CONNECT;
```
MikaelSmith: http://curl.haxx.se/libcurl/c/CURLOPT_URL.html - CURLE_OUT_OF_MEMORY can occur with insufficient heap space.

curl/tests/mock_curl.cc
```diff
@@ -0,0 +1,361 @@
...
+            }
+            h->header = va_arg(vl, curl_slist*);
+            break;
+        case CURLOPT_COOKIE:
+            // Set the mock curl Cookie header to that which was passed in the request.
+            if (h->test_failure_mode == curl_impl::error_mode::set_cookie_error) {
+                va_end(vl);
+                return CURLE_COULDNT_CONNECT;
```
MikaelSmith: http://curl.haxx.se/libcurl/c/CURLOPT_COOKIE.html - CURLE_OUT_OF_MEMORY

curl/tests/mock_curl.cc
```diff
@@ -0,0 +1,361 @@
...
+            }
+            h->cookie = va_arg(vl, char*);
+            break;
+        case CURLOPT_CAINFO:
+            // Set the mock curl Certificate Authority path to that which was passed in the request.
+            if (h->test_failure_mode == curl_impl::error_mode::ca_bundle_error) {
+                va_end(vl);
+                return CURLE_COULDNT_CONNECT;
```
MikaelSmith: http://curl.haxx.se/libcurl/c/CURLOPT_CAINFO.html - CURLE_OUT_OF_MEMORY

curl/tests/mock_curl.cc
```diff
@@ -0,0 +1,361 @@
...
+            }
+            h->cacert = va_arg(vl, char*);
+            break;
+        case CURLOPT_SSLCERT:
+            // Set the mock curl SSL client cert name to that which was passed in the request.
+            if (h->test_failure_mode == curl_impl::error_mode::ssl_cert_error) {
+                va_end(vl);
+                return CURLE_COULDNT_CONNECT;
```
MikaelSmith: http://curl.haxx.se/libcurl/c/CURLOPT_SSLCERT.html - CURLE_OUT_OF_MEMORY

curl/tests/mock_curl.cc
```diff
@@ -0,0 +1,361 @@
...
+            }
+            h->client_cert = va_arg(vl, char*);
+            break;
+        case CURLOPT_SSLKEY:
+            // Set the mock curl private keyfile name to that which was passed in the request.
+            if (h->test_failure_mode == curl_impl::error_mode::ssl_key_error) {
+                va_end(vl);
+                return CURLE_COULDNT_CONNECT;
```
MikaelSmith: http://curl.haxx.se/libcurl/c/CURLOPT_SSLKEY.html - CURLE_OUT_OF_MEMORY

curl/tests/mock_curl.cc
```diff
@@ -0,0 +1,361 @@
...
+
+/*
+ * Perform a cURL transfer. the mock implementation uses this method to
+ * write the response header and body.
+ */
+CURLcode curl_easy_perform(CURL * easy_handle)
+{
+    auto h = static_cast<curl_impl*>(easy_handle);
+    if (h->test_failure_mode == curl_impl::error_mode::easy_perform_error) {
+        return CURLE_COULDNT_CONNECT;
```
MikaelSmith: This seems fair. Way too many error options to check all of them.

curl/tests/mock_curl.cc
```diff
@@ -0,0 +1,361 @@
...
+                                    "Connection: keep-alive\n"
+                                    "Date: Thu, 16 Jul 2015 18:41:08 GMT\n"
+                                    "Content-Type: text/html;charset=UTF-8\n"
+                                    "Server: Jetty(7.x.y-SNAPSHOT)\n"
+                                    "Via: 1.1 vegur";
+
+            h->write_header(&header_content[0], 1, header_content.size(), h->header_context);
+        } else if (h->request_url == "http://nonstd-header.com/") {
+            string header_content = "nonstd_header_name:nonstd_header_value";
+            h->write_header(&header_content[0], 1, header_content.size(), h->header_context);
```
MikaelSmith: I wonder if curl would really return CURLE_OK if it got these headers back. Maybe it doesn't care and just acts as a transport.

curl/tests/mock_curl.cc
```diff
@@ -0,0 +1,361 @@
...
+}
+
+/*
+ * Simply return a nullptr, as error description is not necessary
+ * for testing.
+ */
+const char *curl_easy_strerror(CURLcode errornum)
+{
+    switch (errornum) {
+        case 0:
```
MikaelSmith: Why are these numbers rather than CURL enums?

curl/tests/mock_curl.cc
```diff
@@ -0,0 +1,361 @@
...
+            break;
+        case 2:
+            return "cURL failed init with: CURLE_FAILED_INIT";
+        case 7:
+            return "cURL failed init with: CURLE_COULDNT_CONNECT";
+        default:
+            break;
```
MikaelSmith: You could just `return nullptr` here instead of break.

curl/tests/mock_curl.cc
```diff
@@ -0,0 +1,361 @@
...
+        return new_slist_obj;
+    }
+}
+
+/*
+ * Unimplemented, as we allocate very little memory for curl_slist
+ * objects in tests.
+ */
+void curl_slist_free_all(struct curl_slist * list)
+{
```
MikaelSmith: This may have to change if we ever want to run memory checks on unit tests.

curl/tests/request_test.cc
```diff
@@ -0,0 +1,116 @@
...
+using namespace std;
+
+namespace leatherman { namespace curl {
+
+    struct mock_client : client {
+        curl_handle const& get_handle() { return client::get_handle(); }
+    };
+
+    TEST_CASE("curl::request") {
+        mock_client test_client;
```
MikaelSmith: You don't need a test_client for the first set of test, I'd break them into separate cases. And if you're testing request, just create the request object before the sections.

curl/tests/request_test.cc
```diff
@@ -0,0 +1,116 @@
...
+                expected_value = "header_value_" + to_string(i);
+                REQUIRE(name == expected_name);
+                REQUIRE(value == expected_value);
+                ++i;
+                return true;
+            });
+        }
+
+        SECTION("Cookies should be present in the request when added") {
+            request test_request {"http://valid.com"};
```
MikaelSmith: If you split the sections up, with those using the client in a new TEST_CASE, you could pull a lot of setup for these tests out before the sections as well.

MikaelSmith: There may be other ways to nest tests as well. https://github.com/philsquared/Catch/blob/master/docs/test-cases-and-sections.md

MikaelSmith:
First pass done.

MikaelSmith:
Also I'd rebase to pick up a change to travis config.

curl/tests/mock_curl.cc
whopper: @MikaelSmith good point. Perhaps I'll return a `CURLE_UNKNOWN_OPTION` just Tt hit that `http_request_exception` in the client? Do you think it's alright to return a potentially impossible value to test the exception?

curl/tests/mock_curl.cc
whopper: Doh, will fix that.

curl/tests/mock_curl.cc
MikaelSmith: Yeah, I guess it still makes sense.

whopper:
@MikaelSmith thanks for the review! I realized that before I left I'd forgotten to push up my test refactoring, which I did just now along with updates based on your comments.

curl/tests/response_test.cc
```diff
@@ -0,0 +1,65 @@
...
+using namespace std;
+
+namespace leatherman { namespace curl {
+
+    struct mock_client : client {
+        curl_handle const& get_handle() { return client::get_handle(); }
+    };
+
+    TEST_CASE("curl::response") {
+        mock_client test_client;
```
MikaelSmith: No longer used.

whopper: @MikaelSmith updated

curl/tests/response_test.cc
```diff
@@ -0,0 +1,63 @@
...
+
+namespace leatherman { namespace curl {
+
+    struct mock_client : client {
+        curl_handle const& get_handle() { return client::get_handle(); }
+    };
+
+    TEST_CASE("curl::response") {
+        SECTION("Headers should be addable and retrievable from the response") {
+            response test_response;
```
MikaelSmith: You could still pull `response test_response;` out of every SECTION into the TEST_CASE. CATCH ensures a new object is created for each SECTION.

curl/tests/request_test.cc
MikaelSmith: I feel like these tests that use a client should be in the client file.

curl/tests/request_test.cc
```diff
@@ -0,0 +1,116 @@
...
+
+    struct mock_client : client {
+        curl_handle const& get_handle() { return client::get_handle(); }
+    };
+
+    TEST_CASE("curl::request") {
+        mock_client test_client;
+
+        SECTION("Headers should be addable and retrievable from the request") {
+            request test_request {"http://valid.com"};
```
MikaelSmith: All the `request test_request {"http://valid.com"};` can be pulled out of SECTIONs into the TEST_CASE.

MikaelSmith:
Last 2 knit-picks.

curl/tests/client_test.cc
```diff
@@ -0,0 +1,341 @@
...
+
+        SECTION("Cookies should be present in the request when added") {
+            test_request.add_cookie("cookie_name", "cookie_val");
+            auto resp = test_client.get(test_request);
+            CURL* const& handle = test_client.get_handle();
+            auto test_impl = static_cast<curl_impl* const>(handle);
+            REQUIRE(test_impl->cookie == "cookie_name=cookie_val");
+        }
+
+        SECTION("Cookies should be removable from the request") {
```
whopper: @MikaelSmith I added the two tests from `request_test.cc` which use test clients into the cookie section of the client tests.

curl/tests/response_test.cc
MikaelSmith: And this comment?

whopper: @MikaelSmith good call - updated.

MikaelSmith:
:+1:

PR #58, opened 2015-07-24
(maint) Add option to build dynamic libraries

MikaelSmith:
It's quite likely we'll want to build dynamic libraries of Leatherman
modules. Add that capability as an option to build instead of static
libraries, disabled by default.


MikaelSmith:
For fun, build and install dynamic libraries, download https://root.cern.ch/drupal/content/cling, and run
```
[cling]$ .I /usr/local/include
[cling]$ .L /usr/local/lib/libleatherman_curl.dylib
[cling]$ #include <leatherman/curl/client.hpp>
[cling]$ #include <iostream>
[cling]$ using namespace leatherman::curl;
[cling]$ using namespace std;
[cling]$ client http;
[cling]$ auto r = request("http://www.example.com");
[cling]$ auto resp = http.get(r);
[cling]$ cout << resp.body();
```

cmake/internal.cmake
```diff
@@ -78,10 +78,17 @@ endmacro()
 # This macro cannot be invoked multiple times
 macro(add_leatherman_library)
     include_directories(${${include_var}})
-    add_library(${libname} STATIC ${ARGV})
+    if(LEATHERMAN_SHARED)
+        add_library(${libname} SHARED ${ARGV})
+        target_link_libraries(${libname} ${${deps_var}})
```
MikaelSmith: Might be able to clear `${deps_var}` after doing this. I haven't thought it through.

branan: I wouldn't. Not all platforms will implicitly pull in the dependencies of a shared object at link time, so we still need the deps for the FindLeatherman step (at least sometimes on old/weird unixes)

MikaelSmith:
Updated.

MikaelSmith:
After LTH-61, make which libnowide configurable based on the new flag.

peterhuene:
Does this need to go in before or after PR #107?  Just checking before I merge this.

MikaelSmith:
I can update #107 after it's merged.

PR #60, opened 2015-07-27
(LTH-34) Add new JsonContainer::size method

parisiale:
This method provides the number of entries of object and array entries
or 0 in case of scalars.


MikaelSmith:
Travis seems to fail on an assertion in JsonContainer.

PR #63, opened 2015-07-28
(maint) Minor cleanup of the error reporting in the Ruby API

Magisus:


ruby/src/api.cc
```diff
@@ -469,7 +469,7 @@ namespace leatherman { namespace ruby {
                                                                   "break file if File.exist? file;"
                                                                   "false end)" });
         if (!success) {
-            LOG_WARNING("ruby failed to run", output, none);
+            LOG_WARNING("ruby failed to run");
```
MikaelSmith: Comparing te [stable](https://github.com/puppetlabs/facter/blob/stable/lib/src/ruby/api.cc#L509), this should be `"ruby failed to run: %1%", output`.

ruby/src/api.cc
```diff
@@ -469,7 +469,7 @@ namespace leatherman { namespace ruby {
                                                                   "break file if File.exist? file;"
                                                                   "false end)" });
         if (!success) {
-            LOG_WARNING("ruby failed to run", output, none);
+            LOG_WARNING("ruby failed to run: %1%", output);
```
MikaelSmith: This used to `return library;` after the warning.

MikaelSmith:
Ok, I think that's everything. @peterhuene?

peterhuene:
:+1:

MikaelSmith:
@Magisus can you bump the submodule in Facter#master?

PR #64, opened 2015-07-29
(LTH-33) JsonContainer::get type checks

parisiale:
New error classes.
Checking the type of the requested entry in get methods (behaviour change).
Improving unit tests.
Style improvements.


parisiale:
@briancain we have a behavior change here; in case you call JsonContainer::get<T>(a_key) with the wrong type T, you'll get a new data_type_error, that you can catch, instead of an assertion error from rapidjson.

briancain:
@parisiale Thanks for the info! 

@nicklewis @glennadjrussell @lutter @EGlenfield22 for future info when working with our projects :)

PR #65, opened 2015-07-29
(maint) Add missing <functional> include to strings.hpp

branan:
This is apparently pulled in automatically from other headers on some
compiler versions, but not on my GCC 5.2.0


Magisus:
Seems legit.

PR #66, opened 2015-07-29
(FACT-1123) Add LoadError to ruby API bindings

branan:
The fact-1123 implementation uses a LoadError to signal when
initialization of libfacter has failed. This ports the necessary
changes from facter.


peterhuene:
:+1:

PR #67, opened 2015-07-31
(maint) Set Content-Length header based on size of the body

nicklewis:
When providing a body to a POST request, setting the Content-Length
header to the size of the body is required in order for the server to
properly read the body. Currently that's a manual process that always
needs to be done, in the same way, by the user of the curl client for
every request. Since the behavior is always the same and it's always
necessary, the body setter now computes and adds the header
automatically. This can still be overridden by the consumer by setting
the header before adding the body, if they don't want it set (though
it's hard to imagine a reason for that).

### DIFF
--- a/curl/src/request.cc
+++ b/curl/src/request.cc
@@ -82,6 +82,7 @@ namespace leatherman { namespace curl {
     {
         _body = move(body);
         add_header("Content-Type", move(content_type));
+        add_header("Content-Length", to_string(_body.size()));
     }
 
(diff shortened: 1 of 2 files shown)


MikaelSmith:
:+1:

MikaelSmith:
Actually, I'd like to keep using the commit format we use for public repos; i.e. add `(maint)` to the title.

nicklewis:
Updated commit and PR subjects.

PR #69, opened 2015-08-04
(maint) Add rt dependency for Boost.Thread

MikaelSmith:
On some Linux and Solaris versions, rt must be explicitly linked as a
dependency for libboost_thread. Add it to dependencies in Logging, the
only library that uses Boost.Thread.

### DIFF
--- a/logging/CMakeLists.txt
+++ b/logging/CMakeLists.txt
@@ -7,6 +7,10 @@ add_leatherman_includes("${Boost_INCLUDE_DIRS}")
 leatherman_dependency(nowide)
 leatherman_dependency(locale)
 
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Tested on CentOS 6 in Facter.

Merging this can eliminate https://github.com/puppetlabs/facter/blob/master/lib/CMakeLists.txt#L152 and https://github.com/puppetlabs/facter/blob/master/lib/CMakeLists.txt#L174.

stahnma:
@MikaelSmith does this also mean that https://github.com/puppetlabs/puppet-access/pull/5 wouldn't be needed, as it would pull from leatherman?

MikaelSmith:
That is what it should mean.

MikaelSmith:
@pcarlisle or @stahnma feel free to merge this to unblock the downstream work.

PR #70, opened 2015-08-07
(LTH-39) Port Facter stable execution fix to Leatherman.

peterhuene:
Porting the Facter stable fix to execution that replaced `access` with
an implementation based on `stat` for consistency between platforms.

See FACT-1155 for more information.


MikaelSmith:
:+1:

PR #74, opened 2015-08-18
(LTH-41) Add function to specify curl protocols

briancain:
This commit adds a wrapper to specify CURLOPT_PROTOCOLS per a curl
request. If no protocol is defined, it will default to all protocols. If
a protocol is defined but one is used that wasn't specified (like using
http when https was the only protocol enabled), it will fail with an
http exception saying that an unsupported protocol was used.

### DIFF
--- a/curl/inc/leatherman/curl/client.hpp
+++ b/curl/inc/leatherman/curl/client.hpp
@@ -201,7 +201,7 @@ namespace leatherman { namespace curl {
         std::string _ca_cert;
         std::string _client_cert;
         std::string _client_key;
-        long _client_protocols;
+        long _client_protocols = CURLPROTO_ALL;
 
         response perform(http_method method, request const& req);
         void set_method(context& ctx, http_method method);
--- a/curl/inc/leatherman/curl/client.hpp
+++ b/curl/inc/leatherman/curl/client.hpp
@@ -164,6 +164,13 @@ namespace leatherman { namespace curl {
          */
         void set_client_cert(std::string const& client_cert, std::string const& client_key);
 
+        /**
+         * Set and limit what protocols curl will support
+         * @param client_protocols bitmask of CURLPROTO_*
+         *        (see more: http://curl.haxx.se/libcurl/c/CURLOPT_PROTOCOLS.html)
+         */
+        void set_supported_protocols(long client_protocols);
+
      private:
         client(client const&) = delete;
         client& operator=(client const&) = delete;
@@ -194,6 +201,7 @@ namespace leatherman { namespace curl {
         std::string _ca_cert;
         std::string _client_cert;
         std::string _client_key;
+        long _client_protocols;
 
         response perform(http_method method, request const& req);
         void set_method(context& ctx, http_method method);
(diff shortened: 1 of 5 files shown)


MikaelSmith:
Can you add tests covering this? You'll need to add a clause like https://github.com/puppetlabs/leatherman/blob/master/curl/tests/mock_curl.cc#L209-L216 that saves the protocol in https://github.com/puppetlabs/leatherman/blob/master/curl/tests/mock_curl.hpp#L7.

Tests would probably be something like https://github.com/puppetlabs/leatherman/blob/master/curl/tests/client_test.cc#L190-L203.

MikaelSmith:
Also, I don't see where set_client_protocols is used.

curl/inc/leatherman/curl/client.hpp
```diff
@@ -164,6 +164,13 @@ namespace leatherman { namespace curl {
          */
         void set_client_cert(std::string const& client_cert, std::string const& client_key);
 
+        /**
+         * Set and limit what protocols curl will support
+         * @param client_protocols bitmask of CURLPROTO_*
+         *        (see more: http://curl.haxx.se/libcurl/c/CURLOPT_PROTOCOLS.html)
+         */
+        void set_supported_protocols(long client_protocols);
```
MikaelSmith: Is this expected to be a 32-bit or 64-bit type? We should pretty much never use long, as it has different size on Windows vs other OS.

MikaelSmith: But of course curl declares it as long :-1:

briancain: Yeah, not really sure what to say about this. The function itself is expecting them to be long values, where they are in some form `CURLPROTO_SOMETHING` :(

briancain:
@MikaelSmith `set_client_protocols` is used when `client::perform` is called. The method that sets what protocols for the curl request is `set_supported_protocols` and that is what would be called by the user. So if that method is never called, it will just default to all protocols.

Also yep, I will update the PR to include tests soon.

MikaelSmith:
Oh right, it was hidden away as a 1-liner.

briancain:
@MikaelSmith updated to add a couple of tests.

curl/tests/client_test.cc
```diff
@@ -231,6 +231,14 @@ namespace leatherman { namespace curl {
             auto test_impl = static_cast<curl_impl* const>(handle);
             REQUIRE(test_impl->client_key == "key");
         }
+
+        SECTION("cURL should make an HTTP request with the specified HTTP protocol") {
```
MikaelSmith: Can you also verify CURLPROTO_ALL when set_supported_protocols wasn't called?

curl/src/client.cc
```diff
@@ -312,6 +317,19 @@ namespace leatherman { namespace curl {
         }
     }
 
+    void client::set_client_protocols(context& ctx) {
+        if (!_client_protocols) {
+            // If not set, allow all protocols
+            _client_protocols = CURLPROTO_ALL;
+            LOG_WARNING("No client protocol was set. Allowing all protocols.", _client_protocols);
```
MikaelSmith: The `_client_protocols` argument is unused in this log message. I'd drop the argument.

MikaelSmith:
:+1:

PR #76, opened 2015-08-20
(LTH-43) Add getWithDefault to JsonContainer.

aperiodic:
Before this patch, if users of JsonContainer wanted to try to find
an entry in a JSON object and use a default value if the entry was not
found, they would have to perform a two-step process of checking for the
key and then either getting the key or using the default value. This is
more work than is strictly necessary and leads to superfluous repetition
of this logic. With this patch applied, that logic is centralized in the
`getWithDefault` method, and users can call that function with their
default value instead of implementing the logic themselves.

### DIFF
--- a/json_container/inc/leatherman/json_container/json_container.hpp
+++ b/json_container/inc/leatherman/json_container/json_container.hpp
@@ -246,6 +246,26 @@ namespace leatherman { namespace json_container {
             return getValue<T>(*getValueInJson(keys, true, idx));
         }
 
+        /// Return the value of the specified entry of the root object,
+        /// or default_value if the entry doesn't exist.
+        /// Throw a data_type_error in case the type T doesn't match
+        /// the specified one.
+        template <typename T>
+        T getWithDefault(const JsonContainerKey& key, const T default_value) {
+            auto jval = getValueInJson();
+            auto key_data = key.data();
+
+            if (!isObject(*jval)) {
+                throw data_type_error { "not an object" };
+            }
+
+            if (!hasKey(*jval, key_data)) {
+                return default_value;
+            }
+
+            return getValue<T>(*getValueInJson(*jval, key_data));
+        }
+
         /// Throw a data_key_error in case the root is not a valid JSON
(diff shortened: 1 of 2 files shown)


MikaelSmith:
If we wanted to, we could avoid including `rapidjson/document.h` in the header if we can enumerate all types the template method will need to handle. We'd then be able to move the template definition to the source file.

I don't have a strong reason for doing that, so the PR looks fine as-is.

json_container/inc/leatherman/json_container/json_container.hpp
```diff
@@ -7,17 +7,14 @@
 #include <typeinfo>
 #include <memory>
 
-// Forward declarations for rapidjson
+// The full definition of RapidJSON::GenericValue is needed for the definition
+// of the getWithDefault function in this header.
+#include <rapidjson/document.h>
```
MikaelSmith: This is the first time we require rapidjson headers in the json_container header, so I take it back, this looks intentional.

I think we could preserve only needing forward-declarations by wrapping all the jval manipulation in another function, and if it returns a nullptr then return the default.

json_container/inc/leatherman/json_container/json_container.hpp
```diff
@@ -246,6 +243,25 @@ namespace leatherman { namespace json_container {
             return getValue<T>(*getValueInJson(keys, true, idx));
         }
 
+        /// Return the value of the specified entry of the root object,
+        /// or default_value if the entry doesn't exist.
+        /// Throw a data_type_error in case the type T doesn't match
+        /// the specified one.
+        template <typename T>
+        T getWithDefault(const JsonContainerKey& key, const T default_value) {
```
MikaelSmith: This looks a lot like getValueInJson; just catching the data_key_error and returning a default_value.

json_container/inc/leatherman/json_container/json_container.hpp
parisiale: +1

MikaelSmith:
:+1:

PR #78, opened 2015-08-25
(maint) Merge facter#stable Ruby API changes

MikaelSmith:
In Facter#stable, the Ruby API was updated to include an uninitialize
method for cleaning up Ruby before DLL unload. Move it to Leatherman.


MikaelSmith:
Merges Ruby API changes from https://github.com/puppetlabs/facter/pull/1118.

PR #85, opened 2015-09-14
(LTH-48) Return exit_code from execute command

MikaelSmith:
The execute commands previously had no way to return the exit code from
the child process without throwing an exception. Many use cases include
examining the exit code of a failed child process as normal behavior, so
we should provide a way to get it without using exceptions. Add it to
the return tuple to make it available to callers.

### DIFF
--- a/execution/inc/leatherman/execution/execution.hpp
+++ b/execution/inc/leatherman/execution/execution.hpp
@@ -207,9 +207,9 @@ namespace leatherman { namespace execution {
      * @param file The name or path of the program to execute.
      * @param timeout The timeout, in seconds.  Defaults to no timeout.
      * @param options The execution options.  Defaults to trimming output, merging the environment, and redirecting stderr to null.
-     * @return Returns a tuple of whether or not the command succeeded, output from stdout, and output from stderr (if not redirected).
+     * @return Returns a tuple of whether or not the command succeeded, output from stdout, output from stderr (if not redirected), and the exit code.
      */
-    std::tuple<bool, std::string, std::string> execute(
+    std::tuple<bool, std::string, std::string, int> execute(
         std::string const& file,
         uint32_t timeout = 0,
         lth_util::option_set<execution_options> const& options = { execution_options::trim_output, execution_options::merge_environment, execution_options::redirect_stderr_to_null });
@@ -220,9 +220,9 @@ namespace leatherman { namespace execution {
      * @param arguments The arguments to pass to the program. On Windows they will be quoted as needed for spaces.
      * @param timeout The timeout, in seconds. Defaults to no timeout.
      * @param options The execution options.  Defaults to trimming output, merging the environment, and redirecting stderr to null.
-     * @return Returns a tuple of whether or not the command succeeded, output from stdout, and output from stderr (if not redirected).
+     * @return Returns a tuple of whether or not the command succeeded, output from stdout, output from stderr (if not redirected), and the exit code.
      */
-    std::tuple<bool, std::string, std::string> execute(
+    std::tuple<bool, std::string, std::string, int> execute(
         std::string const& file,
         std::vector<std::string> const& arguments,
         uint32_t timeout = 0,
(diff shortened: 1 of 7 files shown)


MikaelSmith:
ping @peterhuene thoughts on this? We also have a case where it'd be convenient to pass stdin to the child process. That starts to follow a different interface, more like Boost.Process. For the particular stdin use case we don't have any back-and-forth, so we could also pass it all at once and handle it as part of the execute interface.

peterhuene:
It makes sense to always return it.  I think the tuple is getting a bit unwieldy now, though. A structure would make things more clear and one wouldn't have to update the call sites each time we want to return something new, but that is a lot of work to change.

With respect to stdin, taking an additional string parameter, creating a stdin pipe, and asynchronously writing to the pipe until either the entire string is written or the pipe breaks should be fine.  Shouldn't complicate the `select`/`WaitForMultipleObject` logic too much, hopefully.

MikaelSmith:
Yeah, probably don't need to affect `select`/`WaitForMultipleObject` at all. I'm working on some code for it.

peterhuene:
We will as any attempt to write all of the data to stdin may block until the child process reads all of the data, which it might not do until a lot of output occurs, which will in turn block the child waiting for us to read its stdout pipe.

MikaelSmith:
I was considering making a requirement that the child process do all its reading on stdin before writing output, but that may not be a valid assumption.

MikaelSmith:
This is probably good to handle on its own, I'll add a separate ticket/PR for handling input.

execution/src/windows/execution.cc
```diff
@@ -380,7 +380,7 @@ namespace leatherman { namespace execution {
             if (options[execution_options::throw_on_nonzero_exit]) {
                 throw child_exit_exception("child process returned non-zero exit status.", 127, {}, {});
             }
-            return make_tuple(false, "", "");
+            return make_tuple(false, "", "", 127);
```
kylog: Is 127 idiomatic for command not found in Windows land? https://msdn.microsoft.com/en-us/library/windows/desktop/ms681382(v=vs.85).aspx sez "The specified procedure could not be found." which may be the same thing - not sure.

Mostly asking out of curiosity, because being consistent across windows/posix is obviously nice to have.

MikaelSmith: I pulled it from the exception above, but yes it looks like the right type of error.

kylog:
LGTM. And yeah, updating to a struct instead of a quad-tuple makes sense.

MikaelSmith:
Should be fixed up.

MikaelSmith:
Windows failure fixed.

execution/inc/leatherman/execution/execution.hpp
```diff
@@ -187,6 +187,16 @@ namespace leatherman { namespace execution {
     };
 
     /**
+     * Encapsulates return value from executing a process: whether or not the command succeeded, output from stdout, output from stderr (if not redirected), and the exit code.
+     */
+    struct exec_result
```
peterhuene: I'd prefer this called `execution_result` or simply `result` since we're in the `execution` namespace, but can live with `exec_result`.

execution/inc/leatherman/execution/execution.hpp
```diff
@@ -187,6 +187,16 @@ namespace leatherman { namespace execution {
...
+    {
+        bool success;
```
peterhuene: Should probably have doc comment on these members.

execution/inc/leatherman/execution/execution.hpp
```diff
@@ -187,6 +187,16 @@ namespace leatherman { namespace execution {
     };
 
     /**
+     * Encapsulates return value from executing a process: whether or not the command succeeded, output from stdout, output from stderr (if not redirected), and the exit code.
```
peterhuene: Could we move the documentation for the members out of this and into the individual members?

execution/inc/leatherman/execution/execution.hpp
peterhuene: I prefer to have a default constructor when there are members that may not be default initialized.

peterhuene:
Other than above :+1:

PR #86, opened 2015-09-15
(LTH-49) Add child process stdin support for execute

MikaelSmith:
Adds support for passing a string to the `execute` commands with the full
stream to pass to stdin. Input will be interleaved with output as
controlled by the invoked process requesting new input.

Also adds a basic `cat` utility - as one can't be expected on Windows -
with additional options for more complex program behavior.

Also includes LTH-30.


MikaelSmith:
It'd certainly be possible to implement interactive actions with a stream or input callback, but for now I'm keeping it simple.

execution/src/posix/execution.cc
```diff
@@ -188,11 +189,38 @@ namespace leatherman { namespace execution {
                 }
                 pipe.buffer.resize(4096);
             }
+            if (input.first != -1) {
```
MikaelSmith: This can be reworked to share implementation by having a flag in pipe to switch between read/write.

MikaelSmith: Actually, not sure I can. They're different enough that it becomes complicated.

MikaelSmith:
Updated to rebase off LTH-48.

execution/src/posix/execution.cc
```diff
@@ -188,11 +189,38 @@ namespace leatherman { namespace execution {
...
+                }
+            }
             if (max == -1) {
                 // All pipes closed; we're done
                 return;
             }
 
+            if (input.first != -1 && FD_ISSET(input.first, &set)) {
+                // Expecting data on input
+                auto count = write(input.first, input.second.c_str(), input.second.size());
```
branan: Even though this is a blocking write, a signal could still cause less than `size` bytes to be written here. We should loop until everything has been written.

branan: Nevermind, we're inside a while a bunch of lines up.

MikaelSmith: Yeah, there's a mega-loop that tries everything until all the pipes are done.

branan:
Overall this looks good, and the POSIX implementation looks sane. The windows one doesn't have anything that jumps out as wrong, but I'm not super familiar with how pipes work there. Would be nice to get another pair of eyeballs on that

MikaelSmith:
Sounds fair, @joshcooper or @Iristyle?

MikaelSmith:
I'm not sure the POSIX implementation is completely right, it seems to lock up on Mac OS X in one of my tests.

execution/src/posix/execution.cc
```diff
@@ -188,11 +189,38 @@ namespace leatherman { namespace execution {
...
+                if (count < 0) {
+                    if (errno != EINTR) {
+                        LOG_ERROR("stdin pipe write failed: %1%.", format_error());
+                        throw execution_exception("failed to write child input.");
+                    }
+                    // Interrupted by signal
+                    LOG_DEBUG("stdin pipe read was interrupted and will be retried.");
```
branan: s/read/write/

MikaelSmith:
I think the POSIX implementation is now correct. Once a read fails to receive input (because the other end was closed) or a write fails (because the other end was closed or there's no more data to write), the corresponding pipe will be closed. Closing the write pipe for some applications (`cat`) will trigger shutdown.

util/src/windows/scoped_handle.cc
```diff
@@ -0,0 +1,17 @@
+#include <leatherman/util/windows/scoped_handle.hpp>
+#include <windows.h>
+
+using namespace std;
+
+namespace leatherman { namespace util { namespace windows {
+
+    scoped_handle::scoped_handle(HANDLE h) :
+        scoped_resource(h, CloseHandle)
```
peterhuene: It'd be nice if we passed a member function that checked for `INVALID_HANDLE` to avoid needless calls to `Close` with an invalid handle (would probably cause warnings under Application Verifier if we ran it).

execution/src/windows/execution.cc
```diff
@@ -221,17 +222,53 @@ namespace leatherman { namespace execution {
...
 
-    static void read_from_child(DWORD child, array<pipe, 2>& pipes, uint32_t timeout, HANDLE timer)
+    struct input_pipe
+    {
+        scoped_handle handle;
+        string buffer;
+        bool pending;
+    };
+
+    static void rw_from_child(DWORD child, array<pipe, 2>& pipes, input_pipe input, uint32_t timeout, HANDLE timer)
```
MikaelSmith: @joshcooper @peterhuene I could mainly use help confirming the changes to this function make sense. They seem to work, but I might be missing some corner cases.

I'm currently not using OVERLAPPED for the write, and still using a file handle. I think both of those things are ok because the child process isn't going to be attempting to read and write at the same time (probably?)

util/src/windows/scoped_handle.cc
MikaelSmith: Ok, I can look at that. I figured CloseHandle was smart enough to essentially do that check.

peterhuene: `CloseHandle` is smart enough, but tools like AppVerifier might see it as a bug (it'll certainly warn you if you call it with an invalid handle, but I'm not sure if it considers `INVALID_HANDLE_VALUE` to itself be invalid; can't hurt to guard against it though).

execution/src/windows/execution.cc
```diff
@@ -117,11 +118,11 @@ namespace leatherman { namespace execution {
 
         // Format a name for the pipe based on the process and counter
         wstring name = boost::nowide::widen((boost::format("\\\\.\\Pipe\\leatherman.%1%.%2%") %
-            GetCurrentProcessId() %
+            GetCurrentThreadId() %
```
joshcooper: `CreateNamedPipe` will create a pipe that is visible to other processes, which is why they're commonly used for IPC -- the named pipe client connects to a pipe using the same name that the named pipe server used when creating the pipe. So I think we need to include both process and thread id when constructing the name. Also are there are any security issues with allowing other processes to connect to the named pipe?

Can we instead create an anonymous pipe using [`CreatePipe`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365152(v=vs.85).aspx) avoiding the issue altogether?

MikaelSmith: We use NamedPipe specifically for OVERLAPPED I/O. In this case we're also creating both sides of the pipe.

I'm not sure I follow your reasoning why we also need process ID, as thread ID should be unique system-wide.

joshcooper: Ah, I didn't realize overlapped I/O wasn't possible using anonymous pipes, but msdn is pretty clear on that "Asynchronous (overlapped) read and write operations are not supported by anonymous pipes."

Then I guess my only concern is what DACL is applied to the named pipe, since other processes can connect to it. But since we specify 1 as the max instance value it would only be possible between the time the pipe is created and when we connect to it (using `CreateFileW`), and if that latter one fails we raise, so :+1:

MikaelSmith:
I've been working on refactoring the Windows execution, but there are some complicated things going on there. I haven't gotten my refactoring working yet, and I don't think it should hold up getting this PR in.

MikaelSmith:
Tested out in Facter and new code. I'm pretty confident this is ready.

execution/src/windows/execution.cc
```diff
@@ -221,17 +219,71 @@ namespace leatherman { namespace execution {
...
+                    if (count == 0) {
+                        input.handle = {};
+                        break;
+                    }
+
+                    // Register written data
+                    input.buffer.erase(0, count);
+                }
+            }
+
```
MikaelSmith: 5 lines down is a bug, it should be continuing not breaking. That would explain much of the weird behavior I've been running into.

MikaelSmith:
Ok. I could still refactor this to use some inheritance to simplify the execute loops with respect to read/write pipes, but I think I'm out of time before going on vacation. So this should be ready to go.

MikaelSmith:
Darn it cppcheck, that warning is wrong. Using delegating constructors prevents you from doing any other initialization in the list.

MikaelSmith:
Have the start of an experiment with classes at https://github.com/MikaelSmith/leatherman/tree/LTH-49-extra.

execution/src/windows/execution.cc
joshcooper: I figured someone else must have written code like this, and interestingly chromium uses both process and thread id and appends a secure random number. https://chromium.googlesource.com/chromium/src.git/+/master/base/sync_socket_win.cc#46 I imagine the pipe name is predictable otherwise.

Ruby on the other hand uses process id and a serial number (which increments in a non-thread-safeway): https://github.com/ruby/ruby/blob/trunk/win32/win32.c#L6248-L6249. ruby--

Iristyle: I suppose the rand name introduces an (almost negligible) possibility for name collisions (depending on how random it is), where the sequential counter will avoid it.

Process + Thread id makes sense though.  Improvement for the future?

MikaelSmith: I suppose there's a scenario using timeouts where we could release this end of the named pipe but the other end doesn't release. Then this thread shuts down and another with the same thread id is started and reconnects to the existing named pipe. That becomes less likely with both process id and thread id (we could throw in a random number as well just to be safe).

PR #88, opened 2015-09-29
[BLOCKED] (CTH-378) Rotate log files

parisiale:
Enable log file rotation by using the text_file_backend as the backend
sink of boost log. Log file rotation will be performed by file size.

### DIFF
--- a/logging/src/logging.cc
+++ b/logging/src/logging.cc
@@ -79,15 +79,25 @@ namespace leatherman { namespace logging {
         format_record<boost::log::formatting_ostream>(rec, strm);
     }
 
-    void setup_logging(ostream &dst, string locale)
+    template <typename SinkT>
+    static void setup_common_logging_config(boost::shared_ptr<SinkT> sink)
     {
         // Remove existing sinks before adding a new one
         auto core = boost::log::core::get();
         core->remove_all_sinks();
 
+        core->add_sink(sink);
+
+        boost::log::add_common_attributes();
+
+        // Default to the warning level
+        set_level(log_level::warning);
+    }
+
+    void setup_logging(ostream &dst, string locale)
+    {
         using sink_t = sinks::synchronous_sink<color_writer>;
         boost::shared_ptr<sink_t> sink(new sink_t(&dst));
-        core->add_sink(sink);
 
 #if (!defined(__sun) && !defined(_AIX)) || !defined(__GNUC__)
         // Imbue the logging sink with the requested locale.
@@ -95,10 +105,7 @@ namespace leatherman { namespace logging {
         dst.imbue(leatherman::locale::get_locale(locale));
 #endif
 
-        boost::log::add_common_attributes();
-
-        // Default to the warning level
-        set_level(log_level::warning);
+        setup_common_logging_config<sink_t>(sink);
 
         // Set whether or not to use colorization depending if the destination is a tty
         g_colorize = color_supported(dst);
@@ -112,11 +119,7 @@ namespace leatherman { namespace logging {
                        int min_free_space,
                        std::string locale)
     {
-        // Remove existing sinks before adding a new one
-        auto core = boost::log::core::get();
-        core->remove_all_sinks();
-
-        // Add the text file backend, in order to set file rotation
+        // Use the text file as logging backend, in order to set file rotation
         using backend_t = sinks::text_file_backend;
         boost::shared_ptr<backend_t> backend(
             new backend_t(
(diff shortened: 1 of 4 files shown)


logging/src/logging.cc
```diff
@@ -90,6 +104,59 @@ namespace leatherman { namespace logging {
...
+            keywords::target = target.data(),               // dir for rotated files
+            keywords::max_size = max_size,                  // max for stored files, [byte]
+            keywords::min_free_space = min_free_space));    // min drive space, [byte]
+
+        // Scan old files that match the name pattern; consider them already rotated
+        sink->locked_backend()->scan_for_files();
+
+        sink->set_formatter(&color_formatter);
+
+#if (!defined(__sun) && !defined(_AIX)) || !defined(__GNUC__)
```
branan: Would be nice to de-duplicate this (and other) common setup bits from the two logging types

logging/src/posix/logging.cc
```diff
@@ -6,7 +6,8 @@ using namespace std;
 
 namespace leatherman { namespace logging {
 
-    void colorize(ostream& dst, log_level level)
+    template <typename StreamT>
+    void colorize_impl(StreamT& strm, log_level level)
```
branan: Why would we ever colorize something that wasn't an `ostream`? Does boost have streams that don't inherit from the normal std bases?

parisiale: I use that with [this CMake option](https://github.com/puppetlabs/pxp-agent/blob/master/CMakeLists.txt#L12} because I like colors when debugging (mostly other PCP components) while keeping logs on file for future reference.

The `boost::log::formatting_ostream` stream is provided directly to the color_formatter() handler.
The class, defined [here](https://github.com/boostorg/log/blob/master/include/boost/log/utility/formatting_ostream.hpp#L88), is used [here](https://github.com/boostorg/log/blob/master/include/boost/log/utility/formatting_ostream_fwd.hpp#L41) for the type:
```
#ifdef BOOST_LOG_USE_CHAR
typedef basic_formatting_ostream< char > formatting_ostream;
#endif
```

logging/src/logging.cc
parisiale: Yep, that can be done

parisiale: I'm not refactoring the locale imbue() call to avoid having a template dealing with very different types: a std stream and a shared ptr to a boost log sink.

logging/src/logging.cc
```diff
@@ -81,15 +105,56 @@ namespace leatherman { namespace logging {
...
 
+    void setup_logging(std::string file_name,
+                       std::ios_base::openmode open_mode,
+                       std::string target,
+                       int rotation_size,
+                       int max_size,
+                       int min_free_space,
+                       std::string locale)
+    {
+        // Use the text file as logging backend, in order to set file rotation
```
parisiale: Perhaps it's worth calling
```
boost::filesystem::path::imbue(std::locale("C"));
```
as suggested by [boost](http://www.boost.org/doc/libs/1_59_0/libs/log/doc/html/log/rationale/why_crash_on_term.html).
@branan, thoughts?

branan: We currently leave that up to the individual application, rather than doing it automagically for all programs. In facter it's at https://github.com/puppetlabs/facter/blob/master/lib/src/logging/logging.cc#L11

IIRC there was a reason for this, but I don't remember precisely why we did it this way.

logging/src/posix/logging.cc
branan: wow, so formatting_ostream really doesn't inherit from ostream :( That makes me super sad, especially given its name.

Template seems fine here in that case.

logging/src/logging.cc
branan: Although if you're opening a file for logging, imbuing the filesystem makes a bit more sense than forcing it automagically in non-file logging setup...

logging/src/logging.cc
```diff
@@ -81,15 +107,61 @@ namespace leatherman { namespace logging {
...
+                       std::string target,
+                       int rotation_size,
+                       int max_size,
+                       int min_free_space,
+                       std::string locale)
+    {
+        // Initialize locale to default before boost::log in order to avoid
+        // crashes during process termination due to file sinks; refer to
+        // www.boost.org/doc/libs/1_59_0/libs/log/doc/html/log/rationale/why_crash_on_term.html
+        boost::filesystem::path::imbue(locale("C"));
```
parisiale: Please have a look at this.

logging/src/logging.cc
```diff
@@ -81,15 +107,61 @@ namespace leatherman { namespace logging {
...
+                       std::string target,
+                       int rotation_size,
+                       int max_size,
+                       int min_free_space,
+                       std::string locale)
+    {
+        // Initialize locale to default before boost::log in order to avoid
+        // crashes during process termination due to file sinks; refer to
+        // www.boost.org/doc/libs/1_59_0/libs/log/doc/html/log/rationale/why_crash_on_term.html
+        boost::filesystem::path::imbue(std::locale("C"));
```
parisiale: @branan please check this.

parisiale:
@branan let's put this on hold; there's a parallel discussion going on regarding who's in charge of triggering the rotation

parisiale:
We agree to let the log file rotation be triggered by an external signal (SIGUSR2 on POSIX).
Will close this PR.

PR #89, opened 2015-10-08
(LTH-51) Protect against potential named pipe re-use

MikaelSmith:
There's a scenario using timeouts where we could release the invoking
end of a named pipe but the other end doesn't release using
Leatherman.Execution. Then the invoking thread shuts down and another
with the same thread id is started and reconnects to the existing named
pipe.

Use the process id and a random number to make that scenario highly
unlikely. This follows a similar pattern used in Chromium:
https://chromium.googlesource.com/chromium/src.git/+/master/base/sync_socket_win.cc#46

### DIFF
--- a/execution/src/windows/execution.cc
+++ b/execution/src/windows/execution.cc
@@ -9,6 +9,9 @@
 #include <boost/filesystem.hpp>
 #include <boost/algorithm/string.hpp>
 #include <boost/nowide/convert.hpp>
+#include <boost/uuid/uuid.hpp>
+#include <boost/uuid/uuid_generators.hpp>
+#include <boost/uuid/uuid_io.hpp>
 #include <cstdlib>
 #include <cstdio>
 #include <sstream>
@@ -107,7 +110,7 @@ namespace leatherman { namespace execution {
     static tuple<scoped_handle, scoped_handle> CreatePipeThrow()
     {
         static LONG counter = 0;
-        static std::mt19937 rd_gen{std::random_device{}()};
+        static boost::uuids::random_generator rand_uuid;
 
         SECURITY_ATTRIBUTES attributes = {};
         attributes.nLength = sizeof(SECURITY_ATTRIBUTES);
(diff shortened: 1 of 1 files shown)


execution/src/windows/execution.cc
```diff
@@ -106,6 +107,10 @@ namespace leatherman { namespace execution {
     static tuple<scoped_handle, scoped_handle> CreatePipeThrow()
     {
         static LONG counter = 0;
+        static auto rd_gen = []() {
+            std::random_device rd;
```
MikaelSmith: Uses the system's best random number generator to generate a seed for a pseudo-random number generator. random_device is slow, so we only use it to generate the seed.

branan:
We're hitting kernel space twice (potentially? I don't know if PID/TID are cached in the C runtime on windows) and *still* running an RNG. Is it worth it to just use UuidCreate (which returns a type-4 random UUID) and UuidToString and simplifying this?

MikaelSmith:
I'd probably consider TID and the counter sufficient, plus a random number sufficient. UuidCreate could be pretty straight-forward instead of the RNG.

The PID is probably unnecessary once you have a decent random number involved.

branan:
My thinking is basically that this is already windows-specific code, and Windows has a function to generate UUIDs. Therefore we shouldn't re-invite our own unique ID system.

But I am not particularly married to that thought, and I'm fine with doing it ourselves if we think it's useful to have some predictability in the names for debugging (PID/TID especially)

peterhuene:
PID and TID are mapped in user space (in the TIB), so those calls are just accessing that information without a kernel transition.  I agree that having the PID and TID in the object's name is useful.  I'm fine with making a PRNG call to provide some randomness.

MikaelSmith:
I suppose a UUID is meant for this, I'll give it a try.

MikaelSmith:
Updated using boost::uuid. UUIDCreate has error-prone memory management.

I tested UUIDCreate, and it's performance isn't noticeably different from boost::uuid or the original random number generator.

execution/src/windows/execution.cc
```diff
@@ -106,16 +110,24 @@ namespace leatherman { namespace execution {
...
+        // pipe, and a counter to generate multiple pipes for the same process invocation.
+        // A scenario exists using timeouts where we could release the invoking end of a named pipe
+        // but the other end doesn't release. Then the invoking thread shuts down and another with
+        // the same thread id is started and reconnects to the existing named pipe. Use the process
+        // id and a random UUID to make that highly unlikely.
+        wstring name = boost::nowide::widen((boost::format("\\\\.\\Pipe\\leatherman.%1%.%2%.%3%.%4%") %
+            GetCurrentProcessId() %
             GetCurrentThreadId() %
-            InterlockedIncrement(&counter)).str());
+            InterlockedIncrement(&counter) %
```
branan: Do we still need the counter now that we have a better unique ID?

MikaelSmith: Statistically unlikely. We do know we'll open several unique pipes for the same command, however.

MikaelSmith:
Are there still disagreements here?

peterhuene:
:+1: @branan ?

PR #92, opened 2015-10-22
(LTH-53) Fix handling of redirect on uploads with curl

pcarlisle:


MikaelSmith:
The unit test appears to have problems with these changes.

pcarlisle:
oh sweet i ran the wrong test suite

pcarlisle:
ok

curl/src/client.cc
```diff
@@ -261,6 +261,33 @@ namespace leatherman { namespace curl {
...
+        result = curl_easy_setopt(_handle, CURLOPT_SEEKFUNCTION, seek_body);
+        if (result != CURLE_OK) {
+            throw http_request_exception(ctx.req, curl_easy_strerror(result));
+        }
+        result = curl_easy_setopt(_handle, CURLOPT_SEEKDATA, &ctx);
+        if (result != CURLE_OK) {
+            throw http_request_exception(ctx.req, curl_easy_strerror(result));
+        }
+
+        switch (method) {
```
MikaelSmith: This seems like it should be dealt with when you handle set_method.

pcarlisle: I can see the argument both ways, but keeping setting the size of the data with setting the data seemed more coherent to me.

pcarlisle: I'll change it if you feel strongly about it though.

MikaelSmith: I'm thinking it over. These two methods are closely related, but the alternatives are that they have to be called in a certain order, or we merge them into one. I'm ok with it the way it is.

MikaelSmith:
Seems ok. :+1:

PR #94, opened 2015-11-02
(LTH-55) Add execute() overloads for PID and stream redirection

parisiale:
Adding two overloads for exectution::execute() to redirect stderr and
stdout streams to file (callback based) and execute a callback that
receives the PID of the spawn process.

Adding new unit tests for the new functionality.

### DIFF
--- a/execution/inc/leatherman/execution/execution.hpp
+++ b/execution/inc/leatherman/execution/execution.hpp
@@ -307,6 +307,52 @@ namespace leatherman { namespace execution {
         lth_util::option_set<execution_options> const& options = { execution_options::trim_output, execution_options::merge_environment, execution_options::redirect_stderr_to_null });
 
     /**
+     * Executes the given program by calling a specified callback that receives the pid of the program's process.
+     * @param file The name or path of the program to execute.
+     * @param arguments The arguments to pass to the program. On Windows they will be quoted as needed for spaces.
+     * @param input A string to place on stdin for the child process before reading output.
+     * @param environment The environment variables to pass to the child process.
+     * @param pid_callback The callback that is called with the pid of the child process. Defaults to no callback, in which case the pid won't be processed.
+     * @param timeout The timeout, in seconds. Defaults to no timeout.
+     * @param options The execution options.  Defaults to trimming output, merging the environment, and redirecting stderr to null.
+     * @return Returns a result struct.
+     */
+    result execute(
+        std::string const& file,
+        std::vector<std::string> const& arguments,
+        std::string const& input,
+        std::map<std::string, std::string> const& environment,
+        std::function<void(size_t)> pid_callback = nullptr,
+        uint32_t timeout = 0,
+        lth_util::option_set<execution_options> const& options = { execution_options::trim_output, execution_options::merge_environment, execution_options::redirect_stderr_to_null });
+
+    /**
+     * Executes the given program by writing the output of stdout and stderr to specified files.
+     * @param file The name or path of the program to execute.
+     * @param arguments The arguments to pass to the program. On Windows they will be quoted as needed for spaces.
+     * @param input A string to place on stdin for the child process before reading output.
+     * @param out_file The file where the output on stdout will be written.
+     * @param err_file The file where the output on stderr will be written. Defaults to no file, in which case the output on stderr will be buffered and returned in the result struct.
+     * @param environment The environment variables to pass to the child process.
+     * @param pid_callback The callback that is called with the pid of the child process. Defaults to no callback, in which case the pid will not be processed.
+     * @param timeout The timeout, in seconds. Defaults to no timeout.
+     * @param options The execution options.  Defaults to trimming output and merging the environment.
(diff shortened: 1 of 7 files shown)


execution/inc/leatherman/execution/execution.hpp
```diff
@@ -307,6 +307,50 @@ namespace leatherman { namespace execution {
...
+     * @param file The name or path of the program to execute.
+     * @param arguments The arguments to pass to the program. On Windows they will be quoted as needed for spaces.
+     * @param input A string to place on stdin for the child process before reading output.
+     * @param environment The environment variables to pass to the child process.
+     * @param pid_callback The callback that is called with the pid of the child process.
+     * @param timeout The timeout, in seconds. Defaults to no timeout.
+     * @param options The execution options.  Defaults to trimming output, merging the environment, and redirecting stderr to null.
+     * @return Returns a result struct.
+     */
+    result execute(
```
MikaelSmith: What does it do with the output? Worth mentioning for both of the new functions.

execution/src/execution.cc
```diff
@@ -218,7 +221,74 @@ namespace leatherman { namespace execution {
...
+        if (!out_stream.is_open()) {
+            throw execution_exception("failed to open the output file.");
+        }
+
+        if (err_file.empty()) {
+            setup_execute(stderr_callback, actual_options);
+        } else {
+            err_stream.open(err_file.c_str(), std::ios::binary);
+            if (!err_stream.is_open()) {
+                out_stream.close();
```
MikaelSmith: Won't this happen on scope exit with ofstream?

execution/src/execution.cc
```diff
@@ -218,7 +221,74 @@ namespace leatherman { namespace execution {
...
+            }
+
+            stderr_callback = ([&](string& line) {
+                err_stream << line << std::endl;
+                return true;
+            });
+        }
+
+        scope_exit stream_cleaner([&]() {
+            out_stream.close();
```
MikaelSmith: Won't this happen on scope exit with ofstream?

MikaelSmith:
This looks pretty safe, it shouldn't break any existing use. Design looks ok with that goal in mind.

Still probably need to consider redesigning for all the uses we've come up with.

execution/src/execution.cc
parisiale: Yes, it's redundant.

execution/inc/leatherman/execution/execution.hpp
parisiale: Right, will do.

parisiale:
I've updated the documentation and removed the ofstream::close() calls.

execution/inc/leatherman/execution/execution.hpp
MikaelSmith: I don't see the documentation updated. I was thinking the summary at the top should be a little more specific, like "Executes the given program, ignoring all output."

And below, "Executes the given program, redirecting output to specified files."

MikaelSmith:
@peterhuene @branan I'd like a 2nd opinion on this approach.

execution/inc/leatherman/execution/execution.hpp
parisiale: Why "ignoring all output"?
The first overload does not ignore stdout/stderr; it does exactly what the one before does, apart from calling the pid_callback. That's why I didn't add anything to its documentation, apart from the pid_callback arg entry.

parisiale:
I've put the changes made in my previous force push in a separate commit, to make visible the documentation update. On top of that, I've added details to the summary sentence.

execution/inc/leatherman/execution/execution.hpp
MikaelSmith: Ah, ok.

MikaelSmith:
:+1:

execution/src/execution.cc
```diff
@@ -218,7 +220,68 @@ namespace leatherman { namespace execution {
...
+        if (err_file.empty()) {
+            setup_execute(stderr_callback, actual_options);
+        } else {
+            err_stream.open(err_file.c_str(), std::ios::binary);
+            if (!err_stream.is_open()) {
+                throw execution_exception("failed to open the error file.");
+            }
+
+            stderr_callback = ([&](string& line) {
+                err_stream << line << std::endl;
```
peterhuene: Do you want to flush the file on every line (i.e. `endl` vs. `\n`)?

peterhuene:
Other than above and the number of arguments to these functions are getting a bit unwieldy (not a fault of this particular change, though), I'm :+1:

branan:
Given how gross our execute stuff is getting, I'm also :+1: on this. We're definitely getting to a place where we should consider redesigning the execution API, though.

execution/src/execution.cc
parisiale: Yeah, using endl was on purpose when I wrote it.
But on second thought, it's not worth.
I will change and squash the two commits.

parisiale:
Removed std::endl from the ofstream writing calls and squashed the two commits.
This should be ready to be merged.

PR #99, opened 2015-11-07
(LTH-58) Infrastructure for message catalogs (i18n)

MikaelSmith:
Adds i18n support via Boost.Locale. Adds several facilities to make this
easy to use, while supporting safe fall-back on platforms where
Boost.Locale doesn't work (Solaris and AIX).

----
Adds a macro, and uses it in Leatherman, to create a target to generate
a translation template for use with gettext. That file will only be
regenerated when the source changes. New messages will be merged into
the existing translation template.

----
Message catalogs for languages can be generated and kept up-to-date by
setting `LEATHERMAN_LOCALES="lang1;lang2;etc"` during cmake
configuration. This is done via a new build target.

----
Adds the `leatherman::locale::format` helper, to be used in place of
`boost::format`. It translates the string passed in, avoiding the
`boost::locale::format(boost::locale::translate(...))` pattern. It's
implemented via a helper class that allows specifying a locale domain
for string translation.

Use a hash_map to keep track of locales by domain. Translation should
use the locale::translate or locale::format helper, which defaults to
the domain of the current project.

----
Allow setting locale lookup paths for testing in projects using
installed Leatherman. Without this change the location would be
hard-coded into installed Leatherman objects.

Use CMAKE_BINARY_DIR to put all .mo files in the same location in a
build, so tests can easily find all translation files.

----
Adds a LEATHERMAN_ENABLE_LOCALE option to control when gettext and
Boost.Locale should be used. Disabled by default on Solaris and AIX.

Also make whether leatherman::locale::format uses boost::locale::format
or boost::format configurable by the project; they use different
substitution strings, so using boost::locale::format is opt-in. The
boost::format behavior will translate from boost::locale::format
substitution characters so the same string can be used in a project
using the i18n support on platforms that don't can't use Boost.Locale.

----

This provides infrastructure for generating gettext translation files from source code (not perfect, but it'll catch literal strings in boost::format, boost::locale::translate, and our LOG_* patterns), compiling them to binary, and setting the the locale to use them for translation. I've (minimally) tested using it as a submodule in Facter as well.

This still has some rough edges:
- FIXED: It sets the global locale. This is the only way translation with boost::format will work. We should look at using boost::locale::format instead (which requires changing the replacement special characters).
- FIXED: We generally only use one ostream for logging, so we can really only use one locale. That means having per-project translation files won't really work. The tooling I created below assumes it will. That gets more complicated with multiple projects that aren't compiled together but that will load each other. Maybe we can specify multiple encodings, and have each project add theirs to the global locale? There are also utilities to explicitly request a translation based on locale, so we could setup a way to make that per-project (probably more defines in the logging header).
- FIXED: There's no recognition of severity level of messages. That may be something we can add when using xgettext (it adds comments for boost::format when you use the --boost option, that may be possible to specify for each type of LOG_* macro).
- FIXED: Modifiers to a boost::format call will have to be translated explicitly if desired (as done in the various log functions). This will be a manual process of finding all that should be translated and wrapping them in boost::locale::translate.
- FIXED: Add file paths for extracted strings, but fix them to use relative paths to the project root
- WON'T FIX: std::locale in GCC doesn't work on AIX or Solaris. This is a bit of an impediment, and might be worth exploring using gettext directly instead.
- WON'T FIX: It'd also be nice to cleanup finding gettext so you can configure it when calling cmake, rather than requiring you add it to PATH.
- WON'T FIX: Deal with Leatherman locale files when used as a submodule. We generally don't install anything from Leatherman when it's a submodule, and testing won't find the po files.


MikaelSmith:
Ideas for getting AppVeyor working with gettext: http://www.boost.org/doc/libs/1_59_0/libs/locale/doc/html/gettext_for_windows.html or [gnuwin32](http://gnuwin32.sourceforge.net/packages/gettext.htm)

Could also use http://www.boost.org/doc/libs/1_59_0/libs/locale/doc/html/messages_formatting.html#multiple_gettext_domain in Leatherman.Logging, along with a define with the project name, to use project-specific domains for each message.

MikaelSmith:
The only short-coming left is avoiding installing Leatherman .mo files when it's a submodule. That's going to take some re-architecting to msgcat them into the parent project's po files, and isn't needed for installed Leatherman. Is that worth the effort?

README.md
```diff
@@ -113,6 +113,36 @@ CMake helpers. These will be automatically added to your
...
+* `gettext_templates <dir> <glob>`: (re)generates the .pot file from
+  source files matching the glob pattern, and updates any translation
+  (.po) files that already exist. Files are put in `dir`. New languages
+  can be added by configuring cmake with LEATHERMAN\_LOCALES.
+* `gettext_compile <dir> <inst>`: creates a `translation` target (also
+  used by `all`) to generate the binary message catalogs (.mo files) and
+  configure installing them to the specified install location (`inst`).
+
+`LEATHERMAN\_LOCALES` expects a quoted semi-colon separated list, as
+in `LEATHERMAN\_LOCALES="en;fr;ja"`.
```
MikaelSmith: Remove the backslash.

README.md
```diff
@@ -113,6 +113,36 @@ CMake helpers. These will be automatically added to your
...
+
+Normal use of cmake/make should ensure the translation files are up-to-
+date. Translations can be tested by setting the LC\_CTYPE environment
+variable.
+
+To translate strings outside of logging, use the
+`leatherman::locale::translate` and `leatherman::locale::format`
+helpers. Strings passed to the helpers will be extracted to .po files.
+`leatherman::locale::format` is a drop-in replacement for
+`boost::format`.
```
MikaelSmith: It's boost::locale::format, link to http://www.boost.org/doc/libs/1_56_0/libs/locale/doc/html/localized_text_formatting.html.

MikaelSmith: Add back extracting comments, so we can add extra commentary about substitution in the string.

cmake/leatherman.cmake
```diff
@@ -131,6 +131,90 @@ macro(enable_cpplint)
...
+
+        file(READ ${lang_template} FILE_CONTENT)
+        string(REPLACE "charset=CHARSET" "charset=UTF-8"
+            FILE_CONTENT "${FILE_CONTENT}")
+        string(REGEX REPLACE "POT-Creation-Date: [^\\]*" "POT-Creation-Date: "
+            FILE_CONTENT "${FILE_CONTENT}")
+        string(REPLACE "${CMAKE_CURRENT_SOURCE_DIR}/" "" FILE_CONTENT "${FILE_CONTENT}")
+        file(WRITE ${lang_template} "${FILE_CONTENT}")
+
+        file(GLOB existing_files ${TRANSLATION_DIR}/*.po)
```
MikaelSmith: Don't manage .po files? Translation workflow assumes they'll monitor project .pot files and generate updated .po files.

locale/tests/locale.cc
```diff
@@ -0,0 +1,70 @@
...
+
+    GIVEN("leatherman::locale::translate") {
+        THEN("messages should be translated") {
+            REQUIRE(translate(literal, proj) == "demande {1,number}.");
+        }
+    }
+
+    GIVEN("leatherman::locale::format") {
+        THEN("messages should be translated") {
+#if defined(__MAC_OS_X_VERSION_MIN_REQUIRED)
```
MikaelSmith: Why?

locale/inc/leatherman/locale/locale.hpp
```diff
@@ -4,13 +4,91 @@
...
      * @param id The locale ID, defaults to a UTF-8 compatible system default.
+     * @param domain The catalog domain to use for i18n via gettext.
+     * @param paths Search paths for localization files.
+     * @return The locale. If a locale for the specified domain already exists, it returns
+     * the same locale until clear_domain is called for that domain.
+     * Throws boost::locale::conv::conversion_error if the system locale is invalid or
+     * the catalog for the specified language can't be used with the system locale encoding.
      */
-    std::locale get_locale(std::string const& id = "");
+    const std::locale& get_locale(std::string const& id = "", std::string const& domain = PROJECT_NAME, std::vector<std::string> const& paths = {PROJECT_DIR});
```
MikaelSmith: Add a warning about locale ref lifetime (until another project calls get_locale) or change to using an iterator preserving data structure.

MikaelSmith:
Updated based on my comments, and now passes CI.

locale/inc/leatherman/locale/locale.hpp
MikaelSmith: Decided to use std::map instead. It preserves iterators, and should be just as fast on the order of locales we'll be dealing with (dozens).

README.md
```diff
@@ -129,10 +160,14 @@ globally. This can be used as follows:
     leatherman_logging_namespace("logging.namespace")
     ...
 
+Initializing logging via setup\_logging will initialize the global
```
MikaelSmith: Not true.

locale/tests/locale.cc
```diff
@@ -0,0 +1,70 @@
...
+
+    GIVEN("leatherman::locale::translate") {
+        THEN("messages should be translated") {
+            REQUIRE(translate(literal, proj) == "demande {1,number}.");
+        }
+    }
+
+    GIVEN("leatherman::locale::format") {
+        THEN("messages should be translated") {
+#if defined(__MAC_OS_X_VERSION_MIN_REQUIRED) || defined(_WIN32)
```
MikaelSmith: Probably related to boost::locale backend (i.e. we need icu).

cmake/leatherman.cmake
```diff
@@ -131,6 +131,88 @@ macro(enable_cpplint)
...
+# Usage: gettext_templates(dir)
+#
+# Create templates for gettext in `dir` from the source files found in additional specified dirs
+# Creates a custom target `translation`.
+macro(gettext_templates dir)
+    find_program(XGETTEXT_EXE xgettext)
+    find_program(MSGINIT_EXE msginit)
+    find_program(MSGMERGE_EXE msgmerge)
+
+    if (XGETTEXT_EXE AND MSGINIT_EXE AND MSGMERGE_EXE)
```
MikaelSmith: Should we invoke this in Make instead CMake?

locale/src/locale.cc
```diff
@@ -0,0 +1,72 @@
...
+        // The system default locale is set with id == "", except on Windows boost::locale's
+        // generator uses a compatible UTF-8 equivalent. Using boost results in UTF-8 being
+        // the default on all platforms.
+        boost::locale::generator gen;
+#ifdef LEATHERMAN_LOCALE_INSTALL
+        // Setup so we can find installed locales.
+        gen.add_messages_path(LEATHERMAN_LOCALE_INSTALL);
+#endif
+        for (auto&& path : paths) {
+            gen.add_messages_path(path);
```
peterhuene: Why the `&&`?  If this is attempting to move `path`, this will need to be casted to an rvalue again for `add_messages_path`.

MikaelSmith: Generally based on http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3853.htm. The alternative here would be `auto const& path`.

peterhuene: Since `paths` is obviously const and the intent wasn't to move the vector's elements, `auto&` is fine for constant references: the type deduced is `T const`, making it `T const&`, as intended.

peterhuene: Also, re: the above paper, using `auto&&` does make sense to support proxies, but those are rarely used if ever (`vector<bool>` is already largely accepted as a silly place).  Personally I would use `auto&` in these places as it is less confused as to the intent.  If it doesn't build, then the developer knows the iterator is returning a proxy and must therefore use `auto&&` to bind to it.  Hopefully the referenced proposal will be included in the language eventually, making a type declaration moot (i.e. the rule becomes "always omit it").

MikaelSmith: Fair, I'll update my practice. Also raised https://github.com/puppetlabs/cppbestpractices/issues/22.

MikaelSmith:
Updated everything except using make instead of cmake for running xgettext.

MikaelSmith:
This should hold off until we do a Leatherman 0.2.0 release, because it requires some breaking changes for people to adapt to.

locale/inc/leatherman/locale/locale.hpp
```diff
@@ -4,13 +4,91 @@
...
+         * @param args Format arguments.
+         * @return The string generated by translating the format string, then applying the arguments.
+         */
+        template <typename... TArgs>
+        std::string operator()(std::string const& s, TArgs... args)
+        {
+            // Create and apply formatter here, as we want to guarantee the lifetime of the arguments.
+            // boost::locale::format doesn't make copies, and a common gotcha is using temporary arguments
+            // to build up the formatter.
+            boost::locale::format form(boost::locale::translate(s));
```
MikaelSmith: Simplify with http://florianjw.de/en/variadic_templates.html?

cmake/leatherman.cmake
MikaelSmith: Doing this with Make instead of CMake requires a cross-platform option for manipulating the file.

MikaelSmith: I can probably invoke CMake directly with a cmake function.

MikaelSmith: Ok, updated this to be make targets.

locale/tests/locale.cc
MikaelSmith: Mac Boost says its built with icu, and I've written tests explicitly asking for icu. So I'm a bit confused.

Minimal test-case
```
#include <boost/locale.hpp>

int main(int argc, char** argv)
{
    boost::locale::format fmt("{1,number} should be 1,25.");
    fmt % 1.25;
    boost::locale::generator gen;
    std::cout << fmt.str(gen("fr-fr.UTF-8")) << std::endl;
    return 0;
}
```

MikaelSmith:
I believe this work is done, but needs three new follow-up tickets:
- Investigate inconsistent number formatting across platforms. Investigate formatting for languages more thoroughly.
- Add documentation for i18n, based on Boost.Locale.

cmake/leatherman.cmake
```diff
@@ -156,6 +160,94 @@ macro(enable_cpplint)
...
+# Create templates for gettext in `dir` from the source files found in additional specified dirs
+# Creates a custom target `translation`.
+macro(gettext_templates dir)
+    find_program(XGETTEXT_EXE xgettext)
+
+    if (XGETTEXT_EXE)
+        set(TRANSLATION_DIR "${dir}")
+        file(MAKE_DIRECTORY ${TRANSLATION_DIR})
+
+        file(GLOB_RECURSE ALL_PROJECT_SOURCES ${ARGN})
```
branan: This is a duplicate GLOB_RECURSE with the one we do at the top level for cpplint. Does it make sense to just do that glob earlier in the toplevel CMakeLists.txt so wqe can pass the same list of files there and here, instead of globbing?

All of our other projects should have a similar glob for lint, so I don't think that this is undue work for other repos

execution/src/windows/execution.cc
```diff
@@ -564,10 +565,10 @@ namespace leatherman { namespace execution {
         if (use_job_object) {
             hJob = scoped_handle(CreateJobObjectW(nullptr, nullptr));
             if (hJob == NULL) {
-                LOG_ERROR("failed to create job object: %1%.", windows::system_error());
+                LOG_ERROR("failed to create job object: {1}.", system_error());
```
branan: Why does this trip the `windows` namespace from `system_error`, but none below do?

cmake/leatherman.cmake
MikaelSmith: That seems reasonable.

MikaelSmith:
I'll do a rebase too.

execution/src/windows/execution.cc
MikaelSmith: That's probably an error in merging. I'm not sure why it would be different.

locale/inc/leatherman/locale/locale.hpp
```diff
@@ -4,13 +4,85 @@
...
+            (void) std::initializer_list<int>{ ((void)(form % args), 0)... };
```
branan: facemelt

MikaelSmith: It's less complicated than it used to be :)

MikaelSmith:
This needs some testing on SPARC and AIX, I'll kick off some builds.

locale/inc/leatherman/locale/locale.hpp
branan: It took me a minute to sort through what it was doing. Having to explain it to @whopper helped dramatically in my understanding

branan:
I haven't looked at the new tests added, but nothing in the implementaiton or design jumps out as wrong.

locale/inc/leatherman/locale/locale.hpp
MikaelSmith: Oh right. I should've included a link for the technique being used there (mainly using `(expr, 0)` to return 0 after evaluating the expression).

MikaelSmith:
Test build should kick off soon at http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/291/.

MikaelSmith:
Not going to work on AIX or Solaris, so I'll update this to disable on those platforms.

MikaelSmith:
Ok, last commit should
- disable the i18n support by default; one impact is that projects don't need to change anything when picking up these changes to Leatherman
- i18n is enabled by defining LEATHERMAN_I18N. Doing so should only be done on platforms where we can use std::locale and Boost.Locale, i.e. not AIX and Solaris. On those platforms, leave the macro disabled. Projects should then change their formatting substitution characters from `%N%` to `{N}`. `leatherman/locale/locale.hpp` will safely convert the formatting strings to `boost::format` style on platforms where LEATHERMAN_I18N is not defined and fall-back to `boost::format`.

MikaelSmith:
Almost there. Our Solaris VMs have an old version of gettext installed on them that doesn't accept the options we use.

MikaelSmith:
\o/

http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/298/

locale/inc/leatherman/locale/locale.hpp
branan: I'm honestly not sure this technique is much clearer than the recursive method of expanding variadic template parameters, but it does avoid a couple more little stubby template functions

branan:
How do we handle plural forms with this?

MikaelSmith:
http://www.boost.org/doc/libs/master/libs/locale/doc/html/messages_formatting.html#plural_forms

However, I simplified the interface to translate too much, I'll work on fixing that.

MikaelSmith:
Updated to remove a stray comment in locale.hpp.

branan:
I'm totally fine pushing plural/context forms to another PR and ticket. This looks pretty sane to me as-implemented for singular forms (besides maybe needing a `rebase -i` to make the history into a smaller number of digestible chunks)

MikaelSmith:
Rebased to simplify the PR.

locale/inc/leatherman/locale/locale.hpp
kylog: Wow. I might need to stare at that line on a full night's sleep.

kylog:
One more rebase needed?

MikaelSmith:
Rebased.

PR #100, opened 2015-11-09
(maint) Use ctor call instead of empty init list in execute signature

parisiale:
Here we try to fix a compilation issue with the signature of a variant
of execution::execute where the env map is defined by default to an
empty initiialization list; we replace that with an explicit call to the
default ctor.


MikaelSmith:
Oh right, clang.

PR #101, opened 2015-11-09
Bugfixes for using leatherman as a standalone installed library

adrienthebo:
The official recommendation from @branan is to install leatherman to the system and build from there, but some testing logic was assuming that leatherman was being used as a vendor library and some cmake files weren't being installed to system locations. This pull request fixes the needful.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,8 @@ endif()
 if (LEATHERMAN_INSTALL)
     set(CMAKE_FILES
         cmake/cflags.cmake
+        cmake/GetGitRevisionDescription.cmake
+        cmake/GetGitRevisionDescription.cmake.in
         cmake/leatherman.cmake
         cmake/options.cmake
         cmake/leatherman_config.cmake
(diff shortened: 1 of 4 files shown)


MikaelSmith:
I suspect you're not using some of the libraries, as I think Ruby has the same problem with test files.

branan:
`add_leatherman_test` shouldn't need to be wrapped, since it just exports a variable (which will be unused in the `find_package` case)

When `add_library` or `add_executable` is explicitly called it does need to be wrapped, since those source files obviously won't exist in an installed Leatherman.It might be worth adding macros for this, but I don't expect us to be writing lots more code that requires fixture libraries/executables (*most* mocking/stubbing can be done without the need for an extra library, and obviously dynamic_library and execution are special in their testing needs)

MikaelSmith:
Oh right. I'd kind of prefer pulling the `add_leatherman_test` sections out so we don't have more confusion about why it's inconsistent.

branan:
Yeah, I agree. Let's only move what we need to into conditionals. It's cleaner, and it'll help us see what we might want to add helpers for in the future

adrienthebo:
Okay, moved the `add_leatherman_test` statements outside of the conditionals.

MikaelSmith:
:+1:

MikaelSmith:
Nitpick, can you add `(maint)` prefixes to the commits?

adrienthebo:
Commit messages fixed up.

MikaelSmith:
Thanks.

PR #102, opened 2015-11-13
(LTH-57) Avoid unnecessary vector copying in JsonContainer

whopper:
Previously, several members of JsonContainer needlessly passed
vectors by value. This commit updates these members to pass by
reference to prevent unnecessary copying, and adds the const
modifier to vector parameters which shouldn't be altered.

In addition, the getWithDefault member has been updated to use
iterators of a vector of keys, rather than altering a copy and
looping over its remaining elements.

### DIFF
--- a/json_container/inc/leatherman/json_container/json_container.hpp
+++ b/json_container/inc/leatherman/json_container/json_container.hpp
@@ -132,11 +132,8 @@ namespace leatherman { namespace json_container {
         /// Throw a data_key_error in case the specified key is unknown.
         std::string toString(const JsonContainerKey& key) const;
 
-        // TODO(ale): we don't use const for the keys arg on a number
-        // of signatures due to gcc issues; fix this
-
         /// Throw a data_key_error in case the specified key is unknown.
-        std::string toString(std::vector<JsonContainerKey> keys) const;
+        std::string toString(const std::vector<JsonContainerKey>& keys) const;
 
         std::string toPrettyString(size_t left_padding) const;
         std::string toPrettyString() const;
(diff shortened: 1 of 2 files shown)


MikaelSmith:
This looks pretty good. :+1:

nicklewis:
`deployer-client` builds and runs properly with this change

parisiale:
Looks great, but this may break a number of things due to gcc issues.
I don't it will require changing a lot, but that's why we postponed that.
:+1: but please keep in mind that we should test by building PCP components and Pegasus with gcc.

json_container/inc/leatherman/json_container/json_container.hpp
```diff
@@ -279,10 +276,9 @@ namespace leatherman { namespace json_container {
         /// the specified one or in case the parent of the specified
         /// entry is not an object.
         template <typename T>
-        T getWithDefault(std::vector<JsonContainerKey> keys, const T& default_value) const {
+        T getWithDefault(const std::vector<JsonContainerKey>& keys, const T& default_value) const {
             auto key_data = keys.back().data();
-            keys.pop_back();
-            auto jval_obj = getValueInJson(keys);
+            auto jval_obj = getValueInJson(keys.cbegin(), keys.cend()-1);
```
pcarlisle: should this really be -1?

pcarlisle: nevermind I get it

whopper:
I went ahead and did a build of pcp-cpp-client with the updated leatherman in Ubuntu 14.04, (using g++), and the build and tests all succeeded: 

```
100%] Building CXX object lib/tests/CMakeFiles/cpp-pcp-client-unittests.dir/unit/validator/validator_test.cc.o
Linking CXX executable ../../bin/cpp-pcp-client-unittests
make[3]: Leaving directory `/root/cpp-pcp-client/build'
[100%] Built target cpp-pcp-client-unittests
...
1: All tests passed (213 assertions in 22 test cases)
1:
1/1 Test #1: cpp-pcp-client library tests .....   Passed    2.55 sec

100% tests passed, 0 tests failed out of 1
```

whopper:
And pegasus also built successfully, and tests passed:

```
100% tests passed, 0 tests failed out of 1
```

PR #104, opened 2015-11-18
(maint) Remove redundant move invocation in construction of ruby api instance

branan:
The move was causing Clang to emit a warning (which we treat as an error due to -Werror). Removing it makes everything better.


peterhuene:
:+1:

PR #108, opened 2015-12-02
(maint) Fix install of libnowide.dll

MikaelSmith:
Previously `make install` on Windows didn't work because it would look
for libnowide.dll on the wrong location. Fix the lookup location and
ensure Windows tests `make install`.


MikaelSmith:
ping @branan

PR #107, opened 2015-12-01
(LTH-61) Use static libnowide by default

MikaelSmith:
Most of Leatherman uses static libraries. On Windows, libnowide is an
exception that makes packaging Leatherman awkward. Use static linking
instead to make Leatherman libraries consistent.

This may require minor fixes for using libnowide if consuming projects
weren't previously synchronizing when they flush `boost::nowide::cout`
or `cerr`.

PR #109, opened 2015-12-07
(LTH-54) Formalize library installs

MikaelSmith:
Create a `leatherman_install` macro to replace CMake's `install` with
one supporting cross-platform defaults.


whopper:
Ping @branan for review - this seems like a pretty light, sane change, but cmake is still pretty new to me so I wouldn't be able to spot any subtle issues in this -- so calling on your cmake knowledge for a quick review and merge :)

cmake/leatherman.cmake
```diff
@@ -59,6 +59,26 @@ macro(defoption name doc default)
...
+# On Windows shared libraries go in bin, import and archive libraries
+# go in lib. On Linux shared libraries go in lib. Binaries go in bin.
+#
+# Also always drop the prefix; give the target its expected name.
+# We often have binaries and related dynamic libraries, and this
+# simplifies giving them different but related names, such as
+# `facter` and `libfacter`.
+macro(leatherman_install)
+    install(TARGETS ${ARGV}
+        RUNTIME DESTINATION bin
```
branan: I think these RUNTIME, LIBRARY, and ARCHIVE destinations are all the CMake defaults?

cmake/leatherman.cmake
```diff
@@ -147,3 +167,19 @@ function(get_commit_string varname)
...
+
+include(GenerateExportHeader)
+# Usage: symbol_exports(TARGET HEADER)
+#
+# Generate the export header for restricting symbols exported from the library,
+# and configure the compiler. Restricting symbols has several advantages, noted
+# at https://gcc.gnu.org/wiki/Visibility.
+macro(symbol_exports target header)
+    generate_export_header(${target} EXPORT_FILE_NAME "${header}")
+    if (NOT APPLE)
```
branan: Why is this "NOT APPLE"? Do we also need to constrain output stuff on other unices besides linux (IIRC AIX doesn't support visibility?)

MikaelSmith: AIX will safely ignore it. Apple will try to use it, but then fail to export things we need so the libraries are unusable. There were some hangups we haven't figured out.

This worked at https://github.com/puppetlabs/cpp-pcp-client/blob/master/lib/CMakeLists.txt#L41-L50, so it should be safe for all our current platforms.

branan: Makes sense, thanks

cmake/leatherman.cmake
MikaelSmith: Worth checking - I'm not sure import libraries go where we want on Windows.

MikaelSmith: It might be useful to update it to support https://github.com/puppetlabs/facter/pull/1228.

MikaelSmith: If I remove all destinations, there don't appear to be any defaults:
`install TARGETS given no ARCHIVE DESTINATION for static library target "leatherman_logging".`

MikaelSmith:
Depends on https://github.com/puppetlabs/leatherman/pull/107 for shared library installs to work.

MikaelSmith:
Should be fixed up.

cmake/leatherman.cmake
MikaelSmith: I'll add a comment.

MikaelSmith:
ping @branan

PR #110, opened 2015-12-08
(maint) Install cpplint.py when installing leatherman

adrienthebo:
We were omitting the cpplint.py script, which meant that `make cpplint`
would only work with a vendored leatherman install. This commit resolves
this by installing that script during installation.


pcarlisle:
shouldn't a non-vendored install just make you install dependencies separately?

MikaelSmith:
Well, this script is part of the Leatherman repo, so installing it makes sense. :+1:

adrienthebo:
@pcarlisle if cpplint.py is not installed by leatherman, what would be the best way for users to install it?

pcarlisle:
I have it via my package manager but looking at where it's installed it
seems fine.

PR #111, opened 2015-12-11
(maint) Fix GCC 5.2 issues in tests

MikaelSmith:
Previous GCC 5.2 work was in the context of building as a submodule. When
building stand-alone we build tests that weren't previously fixed for GCC
5.2 on Windows. Fix them now.


branan:
:+1: Gonna let tests roll, but this looks fine

PR #114, opened 2015-12-11
(LTH-64) Support LIB_SUFFIX for multilib install

MikaelSmith:
Major RPM distributions define the CMake variable LIB_SUFFIX to determine if libraries should be shipped in /usr/lib64 or /usr/lib. Support the LIB_SUFFIX variable.

Depends on https://github.com/puppetlabs/leatherman/pull/109 for LTH-54.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ include(internal)
 if ("${CMAKE_PROJECT_NAME}" STREQUAL "${PROJECT_NAME}")
     if (NOT CMAKE_BUILD_TYPE)
         message(STATUS "Defaulting to a release build.")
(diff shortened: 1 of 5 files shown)


branan:
We probably need to update our toolchain files to ensure LIB_SUFFIX is unset or the empty string. We don't want to be using lib64 when we build/install as part of the puppet-agent build. I think it makes sense to keep the logic of unsetting LIB_SUFFIX off in our own CMake toolchain files, though.

I'm happy to see this go in once we know our own toolchain files are going to work proerly with it.

MikaelSmith:
Put up a PR for toolchain files, and rebased.

MikaelSmith:
I don't see how LIB_SUFFIX would be accidentally set. It's not inherited from environment variables, so it'd have to be explicitly set on the CMake invocation.

branan:
I had assumed that CMake would set it by default on platforms that use it, but I can't find it referenced anywhere in the CMake 3.4 modules. I'm gonna go ahead and merge this

PR #115, opened 2015-12-15
(maint) Fix symbol_exports for non-identifier targets

MikaelSmith:
If the target name is not a valid C-identifier, generate_export_header
will convert it to one. We now do the same thing on Windows when
defining `lib_<target>_EXPORTS`.

### DIFF
--- a/cmake/leatherman.cmake
+++ b/cmake/leatherman.cmake
@@ -184,6 +184,10 @@ macro(symbol_exports target header)
         add_compiler_export_flags()
     endif()
     if (WIN32)
-        add_definitions("-D${target}_EXPORTS")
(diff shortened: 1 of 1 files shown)


cmake/leatherman.cmake
```diff
@@ -184,6 +184,10 @@ macro(symbol_exports target header)
         add_compiler_export_flags()
     endif()
     if (WIN32)
-        add_definitions("-D${target}_EXPORTS")
+        # If the target name is not a C identifier, it will convert it to one. Fix the define
```
kylog: s/it will/generate_export_header will/  ?? It wasn't clear to me.

MikaelSmith: Fixed.

PR #116, opened 2015-12-15
(maint) Use GCC 5.2 in AppVeyor

MikaelSmith:
Use GCC 5.2 and packages from myget in AppVeyor.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,20 +4,19 @@ install:
   - set
   - git submodule update --init --recursive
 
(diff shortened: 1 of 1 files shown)


MikaelSmith:
I'll ticket this. Tagged released artifacts now imply a specific compiler version, so changes to that will have to be tracked.

MikaelSmith:
This can't be merged yet because the C++14 flag will break several compilers.

MikaelSmith:
TODO: document Windows setup in the README.

MikaelSmith:
It'd be cool to have this, but needs to be expanded to cover both GCC 4.8 and GCC 5.2.

PR #117, opened 2015-12-16
DO NOT MERGE Change semantics of file_util::read.

aperiodic:
Change file_util::read to throw errors rather than return the empty
string when the file doesn't exist, can't be read, or an I/O error
occurred while reading it.

*DO NOT MERGE* I put this up so we can talk about the proposed changes, but it's not yet ready (tests don't compile).

### DIFF
--- a/file_util/inc/leatherman/file_util/file.hpp
+++ b/file_util/inc/leatherman/file_util/file.hpp
@@ -12,6 +12,12 @@
 
 namespace leatherman { namespace file_util {
 
+    class file_error : public std::runtime_error {
+      public:
+        explicit file_error(const char *desc) : runtime_error(desc) {}
+        explicit file_error(const std::string& desc) : runtime_error(desc) {}
+    };
+
     /**
      * Reads each line from the given file.
      * @param path The path to the file to read.
@@ -22,18 +28,13 @@ namespace leatherman { namespace file_util {
 
     /**
      * Reads the entire contents of the given file into a string.
+     * Follows symlinks until a non-symlink path is found.
+     * If the file doesn't exist or can't be opened, a file_error is thrown.
+     * If an I/O error occurs while reading the file, a file_error is thrown.
      * @param path The path of the file to read.
      * @return Returns the file contents as a string.
      */
     std::string read(std::string const& path);
-
-    /**
-     * Reads the entire contents of the given file into a string.
-     * @param path The path of the file to read.
-     * @param contents The returned file contents.
-     * @return Returns true if the contents were read or false if the file is not readable.
-     */
-    bool read(std::string const& path, std::string& contents);
 
     /**
      *@return Returns true if the specified file exists and can
--- a/file_util/src/file.cc
+++ b/file_util/src/file.cc
@@ -27,23 +27,32 @@ namespace boost_file = boost::filesystem;
         return true;
     }
 
-    string read(string const& path) {
-        string contents;
-        if (!read(path, contents)) {
-            return {};
-        }
-        return contents;
-    }
+    // This function attempts to read a file at the given path in text mode and return its contents as
+    // a string. If the file at the path is a symlink, it will be followed until a non-symlink file is
+    // found. If anything goes wrong, a file_error will be thrown.
+    std::string read(const boost_file::path& path) {
+      boost_file::path p(path);
+      while (boost_file::is_symlink(p)) {
+        p = boost_file::read_symlink(p);
+      }
 
-    bool read(string const& path, string& contents) {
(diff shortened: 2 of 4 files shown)


file_util/src/file.cc
```diff
@@ -27,23 +27,32 @@ namespace boost_file = boost::filesystem;
...
-        boost::nowide::ifstream in(path.c_str(), ios::in | ios::binary);
-        ostringstream buffer;
-        if (!in) {
-            return false;
-        }
-        buffer << in.rdbuf();
-        contents = buffer.str();
-        return true;
+      boost::nowide::ifstream file(p.c_str());
+      if (!file.is_open()) {
```
pcarlisle: can just check `!file`?

file_util/src/file.cc
```diff
@@ -27,23 +27,32 @@ namespace boost_file = boost::filesystem;
...
-            return {};
-        }
-        return contents;
-    }
+    // This function attempts to read a file at the given path in text mode and return its contents as
+    // a string. If the file at the path is a symlink, it will be followed until a non-symlink file is
+    // found. If anything goes wrong, a file_error will be thrown.
+    std::string read(const boost_file::path& path) {
+      boost_file::path p(path);
+      while (boost_file::is_symlink(p)) {
```
peterhuene: Is the dereferencing of the symlink necessary?  Opening the symlink should follow, right?

file_util/src/file.cc
```diff
@@ -27,23 +27,32 @@ namespace boost_file = boost::filesystem;
...
+      std::stringstream contents_stream(std::ios_base::out);
+      contents_stream << file.rdbuf();
+
+      if (file.fail()) {
+        auto msg = boost::str(boost::format("An I/O error occurred while reading the file %1%: %2%")
+            % p.string() % strerror(errno));
+        throw file_error(msg);
+      }
+
+      return contents_stream.str();
```
pcarlisle: Need two functions, one that works the way facter expects (returns empty string on failure), and this one. Default is the throwing one, call the other read_nothrow.

file_util/src/file.cc
```diff
@@ -27,23 +27,32 @@ namespace boost_file = boost::filesystem;
...
-        ostringstream buffer;
-        if (!in) {
-            return false;
-        }
-        buffer << in.rdbuf();
-        contents = buffer.str();
-        return true;
+      boost::nowide::ifstream file(p.c_str());
+      if (!file.is_open()) {
+        auto msg = boost::str(boost::format("Unable to open file %1% for reading: %2%")
```
pcarlisle: make sure header for this is included

file_util/src/file.cc
```diff
@@ -27,23 +27,32 @@ namespace boost_file = boost::filesystem;
...
-    bool read(string const& path, string& contents) {
-        boost::nowide::ifstream in(path.c_str(), ios::in | ios::binary);
-        ostringstream buffer;
-        if (!in) {
-            return false;
-        }
-        buffer << in.rdbuf();
-        contents = buffer.str();
-        return true;
+      boost::nowide::ifstream file(p.c_str());
```
pcarlisle: is this c_str call necessary?

pcarlisle: prefer list initialization syntax

file_util/src/file.cc
```diff
@@ -27,23 +27,32 @@ namespace boost_file = boost::filesystem;
...
-        string contents;
-        if (!read(path, contents)) {
-            return {};
-        }
-        return contents;
-    }
+    // This function attempts to read a file at the given path in text mode and return its contents as
+    // a string. If the file at the path is a symlink, it will be followed until a non-symlink file is
+    // found. If anything goes wrong, a file_error will be thrown.
+    std::string read(const boost_file::path& path) {
```
pcarlisle: make possible to specify text or binary mode on windows, default to binary

pcarlisle: allow general ios openmode as optional arg, default to `binary | in`

file_util/inc/leatherman/file_util/file.hpp
```diff
@@ -12,6 +12,12 @@
 
 namespace leatherman { namespace file_util {
 
+    class file_error : public std::runtime_error {
```
pcarlisle: should contain the actual error code, look at `std::error_code`

include file name

file_util/src/file.cc
pcarlisle: nothrow should not use exceptions internally

MikaelSmith:
This hasn't seen activity in awhile. I'm going to close it, please re-open when you expect to work on it again.

PR #121, opened 2015-12-22
(LTH-63) Fix dynamic linking in Leatherman

MikaelSmith:


curl/CMakeLists.txt
```diff
@@ -28,6 +28,10 @@ if (BUILDING_LEATHERMAN)
 endif()
 
 add_leatherman_library(src/client.cc src/request.cc src/response.cc)
+if (WIN32)
+    symbol_exports(leatherman_curl "${CMAKE_CURRENT_LIST_DIR}/inc/leatherman/curl/export.h")
```
branan: Is this something that should be folded into the `add_leatherman_library` macro? Something like `add_leatherman_library(src1 src2 src3 EXPORTS /path/to/export.h`?

MikaelSmith: As an option? That could be convenient, sure. I ran into trouble doing it on other platforms though, so I should probably investigate that more first.

MikaelSmith:
I think it's fixed up now, but going to do a few full packaging tests to check.

MikaelSmith:
Tried el-6-x86_64 builds with standalone Leatherman (static and shared); Windows builds with static and shared libcurl and static and shared Leatherman; and an el-7-x86_64 build of PA-131.

branan:
@MikaelSmith yeah, that's a good plan. Do you want to update the matrix as part of this PR?

MikaelSmith:
Sure.

MikaelSmith:
Ran into yet another Windows DLL issue :/

branan:
Travis and appveyor changes look good. Waiting for the builds to finish

MikaelSmith:
Passed!

PR #122, opened 2015-12-22
(maint) Only use curl test library if we're building curl support

branan:
Previously, we were always attempting to munge the leatherman library
list to switch to our mock curl. this is incorrect if we're not
building curl in the first place.


MikaelSmith:
Makes sense. :+1:

PR #123, opened 2015-12-23
(maint) Fix logging_sinks test for color

MikaelSmith:
The logging_sinks test for colorization was original written when
logging was formatted via a a custom formatter. The test created a
custom sink to consume the formatted text, and verified that it
contained the color strings queried from the log level.

In a rewrite to make colored output work on Windows, logging was changed
to use a custom sink so it can configure the console color when writing
to the output stream. The tests were rewritten, but due to test ordering
all the platforms Leatherman unit tests were run on passed because
logging reported color types with colorization disabled. The tests
should have failed because we replaced the sink that did colorization
with one capturing the output for testing.

Testing with streams is already covered by existing tests, so remove the
test using a custom sink. Leatherman.Logging formatting does not apply
to other custom sinks.


branan:
I think we can accept the loss in coverage here

branan:
Actually, on closer inspection, we lost our coverage of the `operator>>` overload for `log_level`: https://coveralls.io/builds/4541050/source?filename=logging%2Fsrc%2Flogging.cc#L156

I'm not sure that's worth writing a test for, but it wouldn't be too difficult. Up to you

MikaelSmith:
I'll add a test tomorrow.

MikaelSmith:
Still working on the test for `operator>>`.

MikaelSmith:
Updated, should have original coverage again.

MikaelSmith:
There were no line coverage changes, so I think the drop in coverage now is simply from having less code to cover (yay percents).

PR #124, opened 2015-12-23
(LTH-71) Avoid matching leatherman_ruby.dll

MikaelSmith:
When checking whether a Ruby DLL has already been loaded on Windows, we
use a regular expression that also happens to match leatherman_ruby.dll.
Avoid matching it.

A more general solution would be good, but requires more significant
changes.

### DIFF
--- a/dynamic_library/src/windows/dynamic_library.cc
+++ b/dynamic_library/src/windows/dynamic_library.cc
@@ -37,18 +37,21 @@ namespace leatherman { namespace dynamic_library {
 
         boost::regex rx(pattern);
         do {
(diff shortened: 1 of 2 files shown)


MikaelSmith:
My feeling is a more general solution either
a) uses a more restrictive set of DLL names that match DLL patterns we know about (probably by examining RubyInstaller)
b) check for the existence of the rb_intern symbol instead.

MikaelSmith:
@peterhuene @branan any thoughts on this PR?

peterhuene:
:+1:

PR #126, opened 2015-12-28
(maint) Explicitly construct empty maps in execution tests

branan:
On OSX with clang/libc++, trying to default-construct a map with an
empty initializer list fails with an error that the constructor is
explicit.


whopper:
:+1: , tested locally and it works!

PR #129, opened 2016-01-14
(doc) Update changelog for Leatherman 0.3.5 release

MikaelSmith:


MikaelSmith:
@branan @melissa looking to do a release to support Windows Vanagon development

Running leatherman.curl unit tests with shared Leatherman and static curl on Windows is still a problem. Probably doesn't affect how we'll build them in Vanagon, if we want to switch it to shared libraries.

melissa:
:+1: yay!

whopper:
:+1: wooo

PR #130, opened 2016-01-29
(maint) Document JsonContainer ctor getting empty string

parisiale:
JsonContainer(std::string s) throws a data_parse_error in case `s` is an
empty string. Here we explicit that in the docs and add a couple of unit
tests.


MikaelSmith:
:+1:

MikaelSmith:
Kicking Travis.

PR #131, opened 2016-02-01
(LTH-81) Drop dependencies when using shared libraries

MikaelSmith:
When Leatherman is built as a set of shared libraries, all required
dependencies are accounted for in the binary files. However the
LEATHERMAN_LIBS and LEATHERMAN_*_LIBS variables still included dependent
libraries, which could be static libraries. When the dependencies are
static libraries, linking using LEATHERMAN_*_LIBS can result in missing
symbols.

Update Leatherman to leave dependencies out of LEATHERMAN_*_LIBS when
built with shared libraries.

### DIFF
--- a/LeathermanConfig.cmake.in
+++ b/LeathermanConfig.cmake.in
@@ -14,6 +14,8 @@ if (LEATHERMAN_USE_LOCALES AND NOT LEATHERMAN_HAVE_LOCALES)
     message(SEND_ERROR "You requested locale support, but leatherman was built without it")
 endif()
 
+set(LEATHERMAN_SHARED @LEATHERMAN_SHARED@)
+
 debug("Selected components: ${Leatherman_FIND_COMPONENTS}")
 foreach(component ${Leatherman_FIND_COMPONENTS})
     string(TOUPPER "${component}" id_upper)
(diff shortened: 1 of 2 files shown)


MikaelSmith:
This will impact some downstream users, as implicit dependencies they were picking up may no longer be included and will have to be called out explicitly. cpp-pcp-client's use of Boost.Chrono is one I ran into /cc @parisiale

MikaelSmith:
Kicking Travis.

branan:
Have you tested that this behaves as expected on Solaris and Windows?

MikaelSmith:
Not really, I'll kick off some builds.

parisiale:
@MikaelSmith I found that "Boost.Thread uses by default Boost.Chrono", so it should come for free at the moment

MikaelSmith:
Hmm, that only happens when using Boost shared libraries. With static libraries you need to be explicit.

MikaelSmith:
I'm also running into a failure with pxp-agent requiring librt that I don't understand. Still investigating that one.

MikaelSmith:
The pxp-agent PR looks like the last issue. I've kicked off a [build](http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/287/) of a subset of puppet-agent platforms to check for other issues.

MikaelSmith:
Build succeeded on debian 7, el 4, el 6, fedora 21, os x 10.10, sles 10, solaris 10 sparc, solaris 11 i386, and ubuntu 10.04. I suppose I'll check AIX as well.

MikaelSmith:
More PRs for AIX: https://github.com/puppetlabs/cpp-pcp-client/pull/137, https://github.com/puppetlabs/pxp-agent/pull/328

MikaelSmith:
Passed AIX build with the additions above.

MikaelSmith:
http://jenkins-release.delivery.puppetlabs.net/job/vanagon_generic_job/295/ and a manual AIX build all passed.

branan:
@MikaelSmith this needs a rebase. If you can get it cleaned up we can get it in this sprint :)

MikaelSmith:
Rebased.

branan:
Just to be on the safe side I'm waiting for appveyor, then I'll click th ebig green button

PR #133, opened 2016-02-04
(maint) Add pod2man cmake macro to leatherman

ajroetker:
...because sharing is caring.

This commit adds a pod2man cmake macro to leatherman so that multiple
teams can use the macro for their man pages without copying the code
around!

### DIFF
--- a/cmake/pod2man.cmake
+++ b/cmake/pod2man.cmake
@@ -1,0 +1,36 @@
+# Taken from https://github.com/tarantool/tarantool
+
+# Generate man pages of the project by using the POD header
+# written in the tool source code. To use it - include this
+# file in CMakeLists.txt and invoke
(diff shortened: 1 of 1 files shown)


ajroetker:
ping @briancain @adrienthebo @MikaelSmith

MikaelSmith:
:+1:

briancain:
Looks good to me :+1:

ajroetker:
@MikaelSmith sort of related question to this PR, but is there a planned release for leatherman 0.3.6? If so could we slot this in before that time? Or if there isn't a release for leatherman planned, how does one go about scheduling a release?

adrienthebo:
:+1:

MikaelSmith:
We cut releases as-needed.

PR #134, opened 2016-02-04
(doc) Add release instructions to README

MikaelSmith:


MikaelSmith:
ping @branan

branan:
:+1:

ajroetker:
@MikaelSmith Might want to put a note about bumping the version in https://github.com/puppetlabs/leatherman/blob/master/CMakeLists.txt#L2 as well

README.md
```diff
@@ -221,6 +246,13 @@ won't help you, and you'll need to write a lower-level CMake
 file. This README can't cover all the possible situations here, but
 the `nowide` and `catch` CMake files are both solid examples.
 
+## How To Release
+
+1. Update [CHANGELOG.md](CHANGELOG.md) with release notes based on
+``git log `git describe --abbrev=0 --tags`..HEAD``
+1. Update the version in the project declaration of [CMakeLists.txt](CMakeLists.txt)
```
MikaelSmith: That's what this line means @ajroetker. Does it need to be clearer in some way?

ajroetker: Sorry it looks like my comment got cut off. I meant to say "... and how you can mark a release as backward incompatible here https://github.com/puppetlabs/leatherman/blob/master/LeathermanConfigVersion.cmake.in"

MikaelSmith: I haven't really dealt with that, do you have some thoughts on what it should say?

ajroetker: "Note: You can specifically mark how far back your leatherman release is backwards compatible [here](https://github.com/puppetlabs/leatherman/blob/master/LeathermanConfigVersion.cmake.in#L4), for instance, if releasing leatherman `0.0.7` and your release is compatible with `0.0.6` leatherman (i.e. only new features), you could change `PACKAGE_VERSION` to `0.0.6`."
or something like that?

MikaelSmith: That doesn't seem like something we'd do around releasing. That seems more like something you do when using Leatherman, and you'd do it by specifying `find_package(Leatherman 0.0.6)`.

ajroetker: Hmm well then yeah maybe just disregard my comments then until some of that has been firmed up, it was just something to consider :dancer:

PR #137, opened 2016-02-08
(maint) Install pod2man macro alongside other leatherman macros

ajroetker:
This commit adds the pod2man macro to the list of install leatherman
macros for downstream consumers.


MikaelSmith:
Oh yup, forgot about that.

PR #139, opened 2016-02-10
(maint) Prep for leatherman 0.3.7 release.

scotje:
We're needing to use the pod2man macro in puppet-code, so I think we need a new release with AJ's fix.


MikaelSmith:
:+1:

PR #140, opened 2016-02-15
(LTH-84) Fix ruby::to_string when not UTF-8

MikaelSmith:
When calling ruby::to_string on a non-UTF-8 encoding, the binary
representation returned will be interpreted incorrectly by the caller.

### DIFF
--- a/locales/leatherman.pot
+++ b/locales/leatherman.pot
@@ -6,7 +6,7 @@
 #, fuzzy
 msgid ""
 msgstr ""
-"Project-Id-Version: leatherman 0.3.6\n"
+"Project-Id-Version: leatherman 0.3.7\n"
 "Report-Msgid-Bugs-To: docs@puppetlabs.com\n"
 "POT-Creation-Date: \n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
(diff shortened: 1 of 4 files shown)


ruby/src/api.cc
```diff
@@ -235,6 +237,7 @@ namespace leatherman { namespace ruby {
     string api::to_string(VALUE v) const
     {
         v = rb_funcall(v, rb_intern("to_s"), 0);
+        v = rb_str_encode(v, utf8_value("UTF-8"), 0, _nil);
```
MikaelSmith: Should this be whatever locale is configured for the current program? i.e. query `leatherman::locale::get_locale()`.

MikaelSmith:
@peterhuene mind taking a look at this? Fixes [FACT-1341](https://tickets.puppetlabs.com/browse/FACT-1341).

locales/leatherman.pot
```diff
@@ -6,7 +6,7 @@
 #, fuzzy
 msgid ""
 msgstr ""
-"Project-Id-Version: leatherman 0.3.6\n"
+"Project-Id-Version: leatherman 0.3.7\n"
```
branan: Should this happen as part of the final version bump to sip?

MikaelSmith: Yeah, need to do a build to ensure the version's updated.

MikaelSmith: Actually, a build may not catch this because it depends on source files changing. Might have to fix it up to update if the version changes. It has no functional impact that I'm aware of however, so we could also maybe just remove version from the template.

ruby/src/api.cc
peterhuene: I think we should assume UTF-8 internally and if there are any C++ libraries or system API interactions that require changing the encoding we can handle those locally.

peterhuene:
:+1:

locales/leatherman.pot
branan: Alright. I'm not gonna worry about it too much for now.

PR #145, opened 2016-03-01
Install generate_translations.cmake file for internationalization

briancain:
This commit updates leathermans cmake file to install the
generate_translations.cmake file that is required when generating
translation files.


MikaelSmith:
Can you add a (maint) prefix?

briancain:
@MikaelSmith ah yep. Forgot to do that even though I had it on my branch name. Updated the PR!

PR #150, opened 2016-03-14
(LTH-90) Execution with new process group on Windows

parisiale:
Adding a new execution_option to the Execution lib to create child
processes in Windows with a new process group and avoid createing
or terminating a related Job Object.

### DIFF
--- a/execution/inc/leatherman/execution/execution.hpp
+++ b/execution/inc/leatherman/execution/execution.hpp
@@ -57,6 +57,10 @@ namespace leatherman { namespace execution {
          */
         preserve_arguments = (1 << 7),
         /**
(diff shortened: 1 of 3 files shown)


MikaelSmith:
This approach seems reasonable. You maintain a connection while the parent job is running for communication, but the parent can halt without terminating the child. According to testing, TerminateProcess will still cause the child to be terminated as well, but halting the parent via Ctrl-C will leave the child running.

MikaelSmith:
I'd prefer to see this as a LTH ticket.

MikaelSmith:
I'll merge this, others can weigh in on whether there's a more appropriate option later.

PR #151, opened 2016-03-31
(LTH-74) Switch BOOST_STATIC to ON by default

MikaelSmith:
Our packaging builds with static Boost libraries. Switch the Leatherman
default to do the same so developer builds are more consistent with
packaging builds.


MikaelSmith:
This is blocked on rebuilding our Linux Boost binaries with static libraries.

PR #154, opened 2016-04-07
(LTH-91) Switch to using fixed-size ints

MikaelSmith:
Longs have different size on Windows and Linux. Prefer to get long longs,
which are always 64-bits.

### DIFF
--- a/ruby/inc/leatherman/ruby/api.hpp
+++ b/ruby/inc/leatherman/ruby/api.hpp
@@ -44,6 +44,11 @@ namespace leatherman {  namespace ruby {
     /**
      * See MRI documentation.
      */
+    using LONG_LONG = int64_t;
+    using ULONG_LONG = uint64_t;
+    /**
+     * See MRI documentation.
+     */
     typedef uintptr_t ID;
 
     /**
@@ -209,11 +214,11 @@ namespace leatherman {  namespace ruby {
         /**
          * See MRI documentation.
          */
-        VALUE (* const rb_num2ulong)(VALUE);
+        ULONG_LONG (* const rb_num2ull)(VALUE);
         /**
          * See MRI documentation.
          */
-        SIGNED_VALUE (* const rb_num2long)(VALUE);
+        LONG_LONG (* const rb_num2ll)(VALUE);
         /**
          * See MRI documentation.
          */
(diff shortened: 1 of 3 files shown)


ruby/inc/leatherman/ruby/api.hpp
```diff
@@ -209,11 +213,11 @@ namespace leatherman {  namespace ruby {
         /**
          * See MRI documentation.
          */
-        VALUE (* const rb_num2ulong)(VALUE);
+        LONG_LONG (* const rb_num2ull)(VALUE);
```
peterhuene: Should this remain `VALUE` as it is supposed to be unsigned (or have a `ULONG_LONG`)?

MikaelSmith: Missed it, should be `unsigned LONG_LONG`, https://github.com/ruby/ruby/blob/trunk/include/ruby/ruby.h#L724

MikaelSmith:
`long` should pretty much never be used in an application meant to compile on 64-bit architectures, except to deal with system calls and libraries you can't control.

MikaelSmith:
Updated.

ruby/inc/leatherman/ruby/api.hpp
```diff
@@ -44,6 +44,11 @@ namespace leatherman {  namespace ruby {
     /**
      * See MRI documentation.
      */
+    using LONG_LONG = SIGNED_VALUE;
+    using ULONG_LONG = unsigned intptr_t;
```
branan: `uintptr_t`?

ruby/src/api.cc
```diff
@@ -238,7 +238,7 @@ namespace leatherman { namespace ruby {
     {
         v = rb_funcall(v, rb_intern("to_s"), 0);
         v = rb_str_encode(v, utf8_value("UTF-8"), 0, _nil);
-        size_t size = static_cast<size_t>(rb_num2ulong(rb_funcall(v, rb_intern("bytesize"), 0)));
+        size_t size = static_cast<size_t>(rb_num2ull(rb_funcall(v, rb_intern("bytesize"), 0)));
```
branan: since `size_t` and `uintptr_t` are the same everywhere we care about (windows, linux, OSX, and AIX) - would it make since for us to just define our rb_num2ull symbol as returning a size_t and avoid this cast?

MikaelSmith: Actually, I think I'm misreading Ruby's source. rb_num2ll and rb_num2ull always return 64-bit types. I lean towards always using that, even on 32-bit systems... mostly because the signature of rb_num2long changed and I don't want to have to support multiple function signatures.

MikaelSmith:
Updated. Being unable to use ruby.h is terrible.

PR #155, opened 2016-04-14
(maint) Fix 32-bit compilation

MikaelSmith:
A change to using 64-bit integers in the Ruby API assumed size_t would
always be 64-bit. That breaks 32-bit compilation. Restore explicitly
casting to size_t.

Getting a size and expecting it to match size_t is also a common
pattern. Add a helper that does bounds-checking for that cast.

### DIFF
--- a/ruby/src/api.cc
+++ b/ruby/src/api.cc
@@ -25,6 +25,11 @@ namespace leatherman { namespace ruby {
     {
     }
 
+    invalid_conversion::invalid_conversion(string const& message) :
+        runtime_error(message)
+    {
+    }
+
 #define LOAD_SYMBOL(x) x(reinterpret_cast<decltype(x)>(library.find_symbol(#x, true)))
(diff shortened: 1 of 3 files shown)


MikaelSmith:
@branan thoughts on this?

ruby/src/api.cc
```diff
@@ -234,12 +234,20 @@ namespace leatherman { namespace ruby {
         return directories;
     }
 
+    size_t api::num2size_t(VALUE v) const
+    {
+        auto size = rb_num2ull(v);
+        if (size > numeric_limits<size_t>::max()) {
+            throw runtime_error("maximum string size exceeded, requested size was " + to_string(size));
```
branan: If this is gonna be in a generically-named function, it probably shouldn't have "string" in the error message?

MikaelSmith: So true.

MikaelSmith:
Updated.

PR #156, opened 2016-04-15
(maint) Add is_bignum to Ruby API

HAIL9000:
Prior to this commit, the Ruby API in leatherman did not have a
method to identify a bignum. Add this method so that we can
identify bignums in addition to fixednums and doubles in facter.

### DIFF
--- a/ruby/tests/api-test.cc
+++ b/ruby/tests/api-test.cc
@@ -1,5 +1,6 @@
 #include <catch.hpp>
 #include <leatherman/ruby/api.hpp>
+#include <limits>
 
 using namespace std;
 using namespace leatherman::ruby;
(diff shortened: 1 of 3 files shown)


MikaelSmith:
Good, but could use tests.

ruby/tests/api-test.cc
```diff
@@ -49,6 +49,11 @@ TEST_CASE("api::is_*", "[ruby-api]") {
 
         REQUIRE(ruby.is_fixednum(ruby.eval("2")));
         REQUIRE_FALSE(ruby.is_fixednum(ruby.eval("1.5")));
+		REQUIRE(ruby.is_fixednum(ruby.eval("12345678912")));
```
MikaelSmith: That number is a bit big to be a fixednum on some Ruby versions on Windows. I would test something like UINT_MAX instead.

MikaelSmith:
@branan this will hopefully pass CI soon, and we can do a Leatherman 0.5.0 release.

PR #158, opened 2016-04-18
(LTH-93) Support building on Nano Server

MikaelSmith:
CMake adds `-lkernel32` to the list of link libraries by default. However,
doing so with MinGW-w64 introduces an unsatisfied dependency on Nano
Server, causing any executable to fail to run. Explicitly set default link
libraries for all C++ projects to avoid linking kernel32.

Also
- Fix output on WMI error
- Ensure explicit dependencies are first

### DIFF
--- a/cmake/cflags.cmake
+++ b/cmake/cflags.cmake
@@ -66,6 +66,13 @@ if (COVERALLS)
 endif()
 
 if (WIN32)
+    # Update standard link libraries to explicitly exclude kernel32. It isn't necessary, and when compiling with
+    # MinGW makes the executable unusable on Microsoft Nano Server due to including __C_specific_handler.
+    SET(CMAKE_C_STANDARD_LIBRARIES "-luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32"
+        CACHE STRING "Standard C link libraries." FORCE)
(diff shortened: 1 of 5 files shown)


MikaelSmith:
ping @Iristyle

cmake/cflags.cmake
```diff
@@ -66,16 +66,23 @@ if (COVERALLS)
...
 
 if (WIN32)
+    # Update standard link libraries to explicitly exclude kernel32. It isn't necessary, and when compiling with
+    # MinGW makes the executable unusable on Microsoft Nano Server due to including __C_specific_handler.
+    SET(CMAKE_C_STANDARD_LIBRARIES "-luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32"
+        CACHE STRING "Standard C link libraries." FORCE)
+    SET(CMAKE_CXX_STANDARD_LIBRARIES "-luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32"
+        CACHE STRING "Standard C++ link libraries." FORCE)
+
     # We currently support Windows Server 2003, which requires using deprecated APIs.
```
branan: This comment has been wrong since we changed WINVER to 0x0600... do you mind throwing up one more commit on this PR to fix this, since it's a windows-focused PR anyway?

MikaelSmith: Sure.

MikaelSmith: Cleaned up. [FACT-1384](https://tickets.puppetlabs.com/browse/FACT-1384) covers the remaining cleanup that was noted in that comment.

PR #160, opened 2016-04-19
(FACT-1284) Expose pid from execute result; Add rb_last_status_set binding

branan:
These are needed for FACT-1284, in order to expose the process exit status to the ruby API

### DIFF
--- a/ruby/inc/leatherman/ruby/api.hpp
+++ b/ruby/inc/leatherman/ruby/api.hpp
@@ -56,6 +56,12 @@ namespace leatherman {  namespace ruby {
      */
     typedef void * rb_encoding_p;
 
+    #ifdef _WIN32
+        typedef int rb_pid_t;
+    #else
+        typedef pid_t rb_pid_t;
+    #endif
+
     /**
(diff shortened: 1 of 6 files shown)


ruby/inc/leatherman/ruby/api.hpp
```diff
@@ -376,6 +382,11 @@ namespace leatherman {  namespace ruby {
         VALUE (* const rb_require)(char const*);
 
         /**
+         * Intern MRI method. We're being naughty
```
kylog: Gulp. But yes, we are needing an internal method here ....

MikaelSmith:
Any spec tests we could modify? I'm pretty sure we have a couple we could check for zero/non-zero pids.

branan:
@MikaelSmith since we have an explicit constructor in the result struct which requires the pid, we get compiler errors if pid isn't passed in (at least we do on GCC 5.3). That's how I caught the "no executable" cases.

branan:
In fact I'd wager (though I'd have to double-check the spec) that since we have a constructor those default values are meaningless

PR #164, opened 2016-04-20
(LTH-94) Add Boost.Regex to all libs using Leatherman.Logging, prepare 0.6.2 release

MikaelSmith:
On AIX, symbols aren't being resolved from other libraries from dynamic
libraries. In particular, the Boost.Log header files included in
Leatherman.Logging's header uses Boost.Regex, requiring any projects
relying on Leatherman.Logging to link Boost.Regex. Add that dependency.

Leatherman.Execution also had dependencies on Boost.Filesystem that were
being fulfilled from other libraries. Ensure those are called out
explicitly as well, again to resolve runtime failures on AIX.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.6.2] - 2016-04-20
+
+### Fixed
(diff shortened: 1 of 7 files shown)


MikaelSmith:
@underscorgan @branan

branan:
I'm +1 on this. Just waiting for CI

PR #166, opened 2016-04-22
(maint) Use Boost.Log sink's flush property

MikaelSmith:
Use the flushing requirement to let Boost.Log control when the sink is
flushed, rather than flushing on every message. This can potentially speed
up writing, as flushes are expensive. It may result in missing some
messages in the case of a crash however.


MikaelSmith:
This could be an interesting optimization, but not critical to anything.

MikaelSmith:
Well, testing says maybe that's a terrible idea.

PR #167, opened 2016-04-22
(LTH-95) Fix unit tests with shared libraries

MikaelSmith:
When curl is a static library, we can't use the stubbing method used by
the Leatherman.curl tests. For now, skip running those tests when curl
is static. In the near future we'll start building Leatherman against a
curl DLL on Windows and use that for testing.


parisiale:
Replicated the issue with curl unit tests on Win 2012 (lth#master 4347a13) and checked the fix, by building with `-DCURL_STATIC=ON, -DLEATHERMAN_SHARED=ON`.

parisiale:
👍

PR #168, opened 2016-04-22
(LTH-96) Fix translation of log messages

MikaelSmith:
If no substitution arguments are specified in a log message, translation
wasn't invoked. Ensure that now it will be, so all uses of log messages
behave consistently.


MikaelSmith:
@parisiale this should be sufficient to fix translating logs.

@briancain I'm surprised you didn't run into this issue. If you do just `LOG_WARNING("some warning");`, it won't actually translate `some warning`. It would only work before if you had a substitution, like `LOG_WARNING("some warning: {1}", 101);`.

MikaelSmith:
@parisiale fixed all your points, just missing testing.

briancain:
@MikaelSmith hmm, that could be because we don't actually use the logging class in leatherman. In our tools we use the translate functions directly.

MikaelSmith:
This is related to [PCP-371](https://tickets.puppetlabs.com/browse/PCP-371).

MikaelSmith:
Added some testing. I'd like to clean up some of the setup between `locale/tests/locale.cc` and `logging/tests/logging_i18n.cc`.

MikaelSmith:
Tests fail everywhere but my laptop :( work for Monday.

MikaelSmith:
Ok, I think this might finally be fixed up.

parisiale:
Tested on OS X; automatic translation for logging::LOG_XXX() calls has been fixed.
Will be testing a bit more.

MikaelSmith:
I'll rebase this in a bit.

logging/tests/logging_on_message.cc
```diff
@@ -1,62 +1,53 @@
...
         LOG_ERROR("error message");
-        THEN("on_message is called with the message") {
-            REQUIRE(level == log_level::error);
-            REQUIRE(message == "error message");
-        }
+        REQUIRE(level == log_level::error);
+        REQUIRE(message == "error message");
     }
-    GIVEN("a FATAL message to log") {
+    SECTION("a FATEL message is logged to on_message") {
```
parisiale: typo

MikaelSmith:
Fixed typo and rebased. I wish I knew why the test disabled in the last commit wasn't working. I'm starting to suspect a bug in Boost 1.58.

parisiale:
Also checked that .mo files are created without running `cmake` twice.
If you're happy to postpone solving the Windows / gcc issue (old Boost?), we can merge this.
👍

MikaelSmith:
I can't reproduce the CI failures (which were avoided with the last commit) at all. Tried setting up an Ubuntu 12.04 instance with essentially the contents of our Travis scripts, and it still passes there.

PR #169, opened 2016-04-22
(maint) Fix formatting on Maintenance

MikaelSmith:
[skip ci]


MikaelSmith:
I was overly optimistic.

@branan formatting look ok on the Maintenance section? Have any others we should add?

branan:
Looks good to me. I think just the two of us is fine for now :+1:

PR #172, opened 2016-04-28
(maint) Disable fragile test

MikaelSmith:
This seems to fail in just about every CI environment, but has been
difficult to reproduce. Skip the test for now.


MikaelSmith:
@McdonaldSeanp feel free to merge and put up a puppet-agent PR to bump Leatherman versions.

PR #173, opened 2016-05-02
(maint) Skip Leatherman.curl tests on Windows

MikaelSmith:
On Windows, mocking libcurl doesn't work in the same fashion it's been
done for other platforms. The only scenario where it does work is when
using static Leatherman libraries and a shared libcurl library. When both
are shared libraries, leatherman_curl.dll depends directly on
libcurl-4.dll, so introducing a mocking dll has no effect.

For now, skip Leatherman.curl tests when they're expected to fail.


MikaelSmith:
Should verify that the unit tests run the same number of cases as the last branch runs.

MikaelSmith:
This was previously skipping tests on Windows when we could run them, so this seems like a better work-around for now.

PR #175, opened 2016-05-07
Changes COM apartment model for Nano Server

alexpilotti:
Windows Nano Server does not support COINIT_APARTMENTTHREADED.
Switching to COINIT_MULTITHREADED fixes the issue while retaining compatibility
with any supported Windows version.

### DIFF
--- a/windows/src/wmi.cc
+++ b/windows/src/wmi.cc
@@ -38,7 +38,7 @@ namespace leatherman { namespace windows {
     wmi::wmi()
     {
         LOG_DEBUG("initializing WMI");
-        auto hres = CoInitializeEx(0, COINIT_APARTMENTTHREADED);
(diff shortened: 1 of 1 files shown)


MikaelSmith:
I think this can work, but will need a coinciding change to https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/com.rb#L210 as well. We'll also need to review if any other gems we use try to initialize com.

alexpilotti:
@MikaelSmith I already filed a bug on Jira for that: https://tickets.puppetlabs.com/browse/PUP-6278

Thanks!

MikaelSmith:
Something like https://github.com/puppetlabs/leatherman/pull/176 should work without breaking Puppet.

PR #176, opened 2016-05-09
(LTH-98) Use COM MTA for Nano Server

MikaelSmith:
Uses COM multi-threaded (MTA) when single-threaded apartments (STA) aren't
supported. Currently that only occurs on Microsoft Nano Server. This
fixes WMI querios on Nano Server.

Does not change the default to multi-threaded because several providers in
Puppet (user, group, MSI package) use WIN32OLE/COM with STA, and using MTA
in Facter would cause them to fail.

### DIFF
--- a/windows/src/wmi.cc
+++ b/windows/src/wmi.cc
@@ -43,9 +43,17 @@ namespace leatherman { namespace windows {
             if (hres == RPC_E_CHANGED_MODE) {
                 LOG_DEBUG("using prior COM concurrency model");
             } else {
-                throw wmi_exception(format_hresult(_("failed to initialize COM library"), hres));
+                // Retry with multi-threaded, in case we're on Nano Server
+                hres = CoInitializeEx(0, COINIT_MULTITHREADED);
+                if (FAILED(hres)) {
+                    throw wmi_exception(format_hresult(_("failed to initialize COM library"), hres));
+                } else {
+                    LOG_DEBUG("COM single-threaded apartment not supported, using multi-threaded");
+                }
             }
-        } else {
+        }
+
+        if (SUCCEEDED(hres)) {
             _coInit = util::scoped_resource<bool>(true, [](bool b) { CoUninitialize(); });
         }
 
--- a/locales/leatherman.pot
+++ b/locales/leatherman.pot
@@ -376,58 +376,63 @@ msgstr ""
 msgid "using prior COM concurrency model"
 msgstr ""
 
-#: windows/src/wmi.cc:46
+#: windows/src/wmi.cc:49
 msgid "failed to initialize COM library"
 msgstr ""
 
-#: windows/src/wmi.cc:56
-msgid "failed to create IWbemLocator object"
+#. debug
+#: windows/src/wmi.cc:51
+msgid "COM single-threaded apartment not supported, using multi-threaded"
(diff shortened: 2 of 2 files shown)


alexpilotti:
LGTM

MikaelSmith:
@branan thoughts on targeting this for Facter?

MikaelSmith:
@alexpilotti @Iristyle An alternative would be to just call CoInitialize, and rely on it to do a sane default for that platform. I assume this doesn't work with future plans for how COM's initialized in Ruby and Puppet?

branan:
@MikaelSmith I know absolutely nothing about how COM is used in our projects, so have no opinion.

alexpilotti:
CoInitialize sets always STA, not supported on Nano.

IMO the easiest option at the moment to retain the old behavior while adding support for Nano is to set MTA on Nano and STA on the rest.

MikaelSmith:
@Iristyle you mind giving this a look?

Iristyle:
Thanks @MikaelSmith - the only question I have here is around initialization order / what we want as a default.  Assuming Leatherman is the first to attempt an init, we get STA on non-Nano, MTA on Nano.

I think this is sensible, but wonder what the implications are of asking for MTA first instead.

What you've done is closest to existing behavior, so I think is correct - but worth thinking about the alternative for consistency.

MikaelSmith:
My original comment tried to address that. Facter is often loaded first in Puppet, and thus the first to initialize COM. If it does that and initialized with MTA, calling OleInitialize later - as done by the WIN32OLE package - will fail. MTA was my first attempt (ignoring failure if it was because we had STA), and I ran into that problem.

Iristyle:
@MikaelSmith that explanation sounds good to me.

windows/src/wmi.cc
```diff
@@ -43,9 +43,17 @@ namespace leatherman { namespace windows {
...
+                if (FAILED(hres)) {
+                    throw wmi_exception(format_hresult(_("failed to initialize COM library"), hres));
+                } else {
+                    LOG_DEBUG("COM single-threaded apartment not supported, using multi-threaded");
+                }
             }
-        } else {
+        }
+
+        if (SUCCEEDED(hres)) {
```
Iristyle: @MikaelSmith The nested logic here has gotten a little tricky, but I think this logic is OK, and this `SUCCEEDED` is indeed necessary.

`S_FALSE` is returned when COM is already initialized, but is a `SUCCEEDED` status, and must be paired with `CoUninitialize`.  `RPC_E_CHANGED_MODE` however, is a `FAILED` status, and can simply be ignored, without any calls to `CoUninitialize`.

Upon the call to retry as MTA, this thread can bail with a `wmi_exception` on failure, or will simply log a message in either of 2 success cases:

* `S_FALSE`, which means it's the second to init -- and must call `CoUninitialize`
* `S_OK`, which means it's the first to init -- and must call `CoUninitialize`

Overall 👍 though it might be better to make the code a little more self-evident, since  most maintainers won't be as familiar with the rules.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms695279(v=vs.85).aspx
https://msdn.microsoft.com/en-us/library/windows/desktop/ff485842(v=vs.85).aspx

PR #178, opened 2016-05-16
(LTH-97) Allow applications to disable locale support

branan:
Previously, if the logging library was built with locale support it
would always be used. This adds a new parameter to the setup_logging
function to allow applications to disable locale support if they
desire.

### DIFF
--- a/locale/src/locale.cc
+++ b/locale/src/locale.cc
@@ -51,7 +51,11 @@ namespace leatherman { namespace locale {
 
     string translate(string const& s, string const& domain)
     {
-        return boost::locale::translate(s).str(get_locale("", domain));
+        try {
+            return boost::locale::translate(s).str(get_locale("", domain));
+        } catch (exception const&) {
+            return s;
+        }
     }
 
 }}  // namespace leatherman::locale
--- a/logging/inc/leatherman/logging/logging.hpp
+++ b/logging/inc/leatherman/logging/logging.hpp
@@ -151,8 +151,9 @@ namespace leatherman { namespace logging {
      * @param dst Destination stream for logging output.
      * @param locale The locale identifier to use for logging.
      * @param domain The catalog domain to use for i18n via gettext.
+     * @param use_locale Whether to use locales in logging setup. If locales are disabled this parameter is ignored.
      */
-    void setup_logging(std::ostream &dst, std::string locale = "", std::string domain = PROJECT_NAME);
+    void setup_logging(std::ostream &dst, std::string locale = "", std::string domain = PROJECT_NAME, bool use_locale = true);
 
     /**
      * Sets the current log level.
--- a/logging/src/logging.cc
+++ b/logging/src/logging.cc
@@ -67,7 +67,7 @@ namespace leatherman { namespace logging {
         _dst << endl;
     }
 
-    void setup_logging(ostream &dst, string locale, string domain)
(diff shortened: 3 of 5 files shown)


locale/src/locale.cc
```diff
@@ -51,7 +51,11 @@ namespace leatherman { namespace locale {
 
     string translate(string const& s, string const& domain)
     {
-        return boost::locale::translate(s).str(get_locale("", domain));
+        try {
+            return boost::locale::translate(s).str(get_locale("", domain));
+        } catch (exception const&) {
+            return s;
```
branan: My reasoning here is:

* If locales are broken we should still return *something* from this function, instead of panicking
* Locale init needs to be done before we can log, so either a) we can't log, or b) the application has initialized logging, and had an opportunity to report to the user that localization is busted

branan:
@MikaelSmith I'd loike your thoughts on this

locale/src/locale.cc
MikaelSmith: Yup, makes sense. Returning the original string is the right thing to do when locales aren't working.

logging/inc/leatherman/logging/logging.hpp
```diff
@@ -151,8 +151,9 @@ namespace leatherman { namespace logging {
      * @param dst Destination stream for logging output.
      * @param locale The locale identifier to use for logging.
      * @param domain The catalog domain to use for i18n via gettext.
+     * @param use_locale Whether to use locales in logging setup. If locales are disabled this parameter is ignored.
      */
-    void setup_logging(std::ostream &dst, std::string locale = "", std::string domain = PROJECT_NAME);
+    void setup_logging(std::ostream &dst, std::string locale = "", std::string domain = PROJECT_NAME, bool use_locale=true);
```
MikaelSmith: It'd be nice to keep spacing consistent.

MikaelSmith:
👍

locale/src/locale.cc
branan: I also think the exception handling overhead is small enough when we're doing translation work that it's not worth memoizing "we can't translate" and making this function more complicated

logging/src/logging.cc
```diff
@@ -85,7 +86,9 @@ namespace leatherman { namespace logging {
         // Note that this creates a locale that's not usable for testing, as it
         // only includes paths for install locations. This is intentional, to avoid leaving
         // searching paths that have unknown permissions.
-        dst.imbue(lth_locale::get_locale(locale, domain, {}));
+        if (use_locale) {
+            dst.imbue(lth_locale::get_locale(locale, domain, {}));
```
MikaelSmith: Imbue here has no real impact. The only reason this line still exists is to call get_locale early, so we can setup the locale for the current project. We're already down this road, but it might've been better to just pull it out as a separate init step for projects.

branan: I think a good "leatherman breaking change" would be disentangling locale and logging setup, since (in theory) an app might want to set up locales for translation without doing any logging at all

locale/src/locale.cc
MikaelSmith: If we were going to do something with memoizing, I'd switch back to directly accessing g_locales, and just return the untranslated string if the locale hadn't been initialized yet. That ends up masking whether we tried to use translate before initializing locales vs locales being busted, but it would be a graceful way to fail.

PR #179, opened 2016-05-17
(LTH-83) Add context and plural support to Leatherman.Locale

MikaelSmith:
Adds context and plural support, based on the pattern from Boost.Locale,
to the Leatherman.Locale `translate` functions.


MikaelSmith:
@parisiale can you review this tomorrow?

parisiale:
Sorry for being late; I missed the notification :-P

Looks good 👍  I got only a minor question, then we're ready to merge. Is it worth to document in the docstrings what happens in case of translation failure (boost::locale exceptions are filtered and the appropriate string, based on `n == 1`, is returned)? I guess it depends on what we're doing elsewhere, I'm not sure.

Tested with cpp-pcp-client & pcp-test.

MikaelSmith:
Updated.

PR #181, opened 2016-05-20
(PCP-426) Add Util::Timer::elapsed_milliseconds and unit tests

parisiale:


util/tests/timer.cc
```diff
@@ -0,0 +1,41 @@
...
+        Timer t {};
+        auto d1_s = t.elapsed_seconds();
+        auto d2_s = t.elapsed_seconds();
+
+        REQUIRE(d1_s <= d2_s);
+    }
+
+    SECTION("can retrieve durations [s]") {
+        Timer t {};
+        auto d1_ms = t.elapsed_milliseconds();
```
MikaelSmith: Should we sleep and then verify `d1_ms < d2_ms`? Seems like a more concrete test.

MikaelSmith:
👍

util/tests/timer.cc
```diff
@@ -0,0 +1,41 @@
...
+
+    SECTION("can retrieve durations [s]") {
+        Timer t {};
+        auto d1_s = t.elapsed_seconds();
+        auto d2_s = t.elapsed_seconds();
+
+        REQUIRE(d1_s <= d2_s);
+    }
+
+    SECTION("can retrieve durations [s]") {
```
parisiale: ms

util/tests/timer.cc
parisiale: Ok, I'll add short pauses, otherwise:
```
/Users/alessandro/code/leatherman/util/tests/timer.cc:29: FAILED:
  REQUIRE( d1_ms < d2_ms )
with expansion:
  0 < 0
```

MikaelSmith: Sleeping 1-2 milliseconds seems like it'd be sufficient. If that brings in significant dependencies, I'd be ok ignoring it.

PR #189, opened 2016-06-16
(LTH-103) Remove line number from .pot files

MikaelSmith:
Including line numbers in gettext .pot files makes it time-consuming to
merge changes, and adds a lot of noise to commits. Remove them.

Also update Puppet Labs to Puppet.


parisiale:
Message changes become more evident. Line numbers in .pot files are not important.
👍

PR #191, opened 2016-06-21
(LTH-99) Fix for Ruby 2.3

MikaelSmith:
Adds aliasing for a symbol that was renamed in Ruby 2.3 (as a
Ruby API deprecation).

### DIFF
--- a/ruby/src/api.cc
+++ b/ruby/src/api.cc
@@ -86,7 +86,7 @@ namespace leatherman { namespace ruby {
         LOAD_SYMBOL(rb_to_id),
         LOAD_SYMBOL(rb_id2name),
         LOAD_SYMBOL(rb_define_alloc_func),
-        LOAD_SYMBOL(rb_data_object_alloc),
+        LOAD_ALIASED_SYMBOL(rb_data_object_alloc, rb_data_object_wrap),
         LOAD_SYMBOL(rb_gc_mark),
         LOAD_SYMBOL(rb_yield_values),
         LOAD_SYMBOL(rb_require),


branan:
@MikaelSmith everything failed on generating message catalogs. Any thoughts?

MikaelSmith:
I added an option in .pot generation that seems to be supported only in very recent versions of gettext. It passed Travis then, probably because regenerating that file depends on git's sync order.

You could probably rerun them to get them to pass. We also need to build a recent gettext version for Travis and make sure CI always rebuilds .pot files.

MikaelSmith:
AppVeyor might be the same issue, haven't looked into it. I haven't seen it fail before for this, only Travis.

MikaelSmith:
CI issue should be fixed.

ruby/src/api.cc
```diff
@@ -86,7 +86,7 @@ namespace leatherman { namespace ruby {
         LOAD_SYMBOL(rb_to_id),
         LOAD_SYMBOL(rb_id2name),
         LOAD_SYMBOL(rb_define_alloc_func),
-        LOAD_SYMBOL(rb_data_object_alloc),
+        LOAD_ALIASED_SYMBOL(rb_data_object_alloc, rb_data_object_wrap),
```
Iristyle: I wonder why they replaced `rb_data_object_alloc` with a method of the same signature?

For ref, Ruby deprecation in 2.3.0 appears at:
https://github.com/ruby/ruby/blob/v2_3_0/include/ruby/ruby.h#L1418-L1425

👍

branan: `_wrap` is definitely a better name, but I dunno why they had to make that change in a Y release -_-

PR #199, opened 2016-07-27
(LTH-107) Add inherit_locale option to execute

MikaelSmith:
The `execute` function was originally implemented for Facter calls to
command-line tools, where it wants consistently parseable output. For
that purpose, it always overrode LC_ALL and LANG.

We now use `execute` for other purposes, so allow inheriting the locale
from the executing process.

### DIFF
--- a/execution/src/posix/execution.cc
+++ b/execution/src/posix/execution.cc
@@ -341,15 +341,15 @@ namespace leatherman { namespace execution {
 
     // Helper function that creates a vector of environment variables in the format of key=value
     // Also handles merging of environment and defaulting LC_ALL and LANG to C
-    static vector<string> create_environment(map<string, string> const* environment, bool merge)
+    static vector<string> create_environment(map<string, string> const* environment, bool merge, bool inherit)
     {
         vector<string> result;
 
         // Merge in our current environment, if requested
         if (merge && environ) {
             for (auto var = environ; *var; ++var) {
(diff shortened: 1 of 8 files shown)


MikaelSmith:
@branan thoughts on this enhancement?

execution/src/posix/execution.cc
```diff
@@ -364,11 +364,12 @@ namespace leatherman { namespace execution {
         }
 
         // Set the locale to C unless specified in the given environment
+        string locale_env;
         if (!environment || environment->count("LC_ALL") == 0) {
-            result.emplace_back("LC_ALL=C");
+            result.emplace_back("LC_ALL="+((inherit && environment::get("LC_ALL", locale_env)) ? locale_env : "C"));
```
MikaelSmith: This isn't quite the behavior we need. If LC_ALL is unset in the parent environment, we should leave it unset in the child process, not set it to C.

MikaelSmith:
Updated, hopefully it works on Windows now.

parisiale:
@MikaelSmith I'm too tired for this now; I'll have a better look in the next few days if @branan doesn't merge if before.

parisiale:
I successfully triggered translated log messages on cpp-pcp-client after setting a different locale (using LTH-107 for cpp-pcp-client & pxp-agent).

PR #200, opened 2016-08-01
(LTH-108) Fix compilation with curl 7.50.0

MikaelSmith:
Makes curl_impl an opaque blob using reinterpret_cast to fix compilation
against curl 7.50.0.


kylog:
Looks like this was needed because of: https://github.com/curl/curl/commit/9adf3c473a01b289c781aab111f9ad2fc541ed4e.

PR #201, opened 2016-08-03
(LTH-109) Add translation helper functions and plural format support

ahenroid:
Add leatherman::locale::format_* for standard, plural, and context-based translations and convenience functions with short names: \_(), n\_(), p_(), np_(). Rename context-based functions from *_c to *_p to align with the naming for these functions/macros in gettext and other i18n libraries (here "p" stands for "prefixed" context).

Also, update xgettext string extraction (generation of .pot files), extend locale translation tests, and remove the helper class leatherman::locale::formatter to improve readability, moving this code out to the respective format_* functions.


locale/inc/leatherman/locale/locale.hpp
```diff
@@ -15,12 +15,11 @@
 #include <locale>
 #include <vector>
 #include <functional>
-
-#ifdef LEATHERMAN_I18N
```
MikaelSmith: I'm not sure why you changed this. Simply including boost::locale on some platforms will be a problem; I don't think we can link against it on one of them (maybe AIX).

MikaelSmith:
Somethings failing when locales are disabled (one of the Travis jobs). Not sure why.

locale/inc/leatherman/locale/locale.hpp
ahenroid: Yes, I realized this and will fix. My initial motivation was to have less conditionally compiled code in this file--magical stuff that happens when LEATHERMAN_I18N is on or off--but obviously this needs to work correctly without boost::locale.

ahenroid:
Ah, I solved the build issue. When locales are disabled, stubs for the translate_* functions are built from locale/disabled/locale.cc.

locale/inc/leatherman/locale/locale.hpp
```diff
@@ -95,63 +95,197 @@ namespace leatherman { namespace locale {
...
+     */
+    template <typename... TArgs>
+    std::string format(std::string const& s, TArgs... args)
+    {
+        // Create and apply formatter here, as we want to guarantee the lifetime of the arguments.
+        // boost::locale::format doesn't make copies, and a common gotcha is using temporary arguments
+        // to build up the formatter.
+        // Technique for the one-liner explained at http://florianjw.de/en/variadic_templates.html.
+#ifdef LEATHERMAN_I18N
+        static const std::string domain{PROJECT_NAME};
```
MikaelSmith: I wish this common code could be collapsed somehow; there are potentially a lot of static function members as well. Maybe the optimizer will handle that well.

One option would probably be to create a very templatized function, and pass in a lambda for the translate/regex_replace calls for each format*.

MikaelSmith:
I think this looks correct 👍

locale/inc/leatherman/locale/locale.hpp
ahenroid: Yes, I was bothered by the replication as well. Let me take a look at your lambda-based approach and alternatives.

ahenroid:
The last commit trimmed down locale.hpp significantly.

locale/inc/leatherman/locale/locale.hpp
```diff
@@ -95,63 +95,161 @@ namespace leatherman { namespace locale {
...
     /**
-     * A formatter class for translating and formatting text according to a domain.
+     * Translates and formats text using the locale initialized by this library.
+     * @param trans The translation function.
+     * @param args Format arguments.
+     * @return The string generated by translating the format string, then applying the arguments.
      */
-    class formatter
+    template <typename... TArgs>
+    std::string _format_common(std::function<std::string(const std::string&)>&& trans, TArgs... args)
```
MikaelSmith: One technique to look up here is anonymous namespaces. Essentially makes the function private to this namespace.

ahenroid:
I moved format_common into a private namespace. Let me know if there's anything else I should address.

MikaelSmith:
The readme might need an update. Otherwise looks good.

PR #202, opened 2016-08-12
(LTH-59) Review and mark translatable strings in Leatherman

ahenroid:
This update marks a variety of exception and error messages for
translation and updates the leatherman.pot file. All logging
messages appear to be properly marked already.

### DIFF
--- a/execution/src/posix/execution.cc
+++ b/execution/src/posix/execution.cc
@@ -3,6 +3,7 @@
 #include <leatherman/util/scope_exit.hpp>
 #include <leatherman/util/posix/scoped_descriptor.hpp>
 #include <leatherman/logging/logging.hpp>
+#include <leatherman/locale/locale.hpp>
 #include <boost/algorithm/string.hpp>
 #include <boost/filesystem.hpp>
 #include <array>
@@ -14,6 +15,9 @@
 #include <fcntl.h>
 #include <signal.h>
 
+// Mark string for translation (alias for leatherman::locale::format)
+using leatherman::locale::_;
+
 using namespace std;
 using namespace leatherman::util;
 using namespace leatherman::util::posix;
@@ -21,7 +25,6 @@ using namespace leatherman::execution;
 using namespace leatherman::logging;
 using namespace leatherman::file_util;
 using namespace boost::filesystem;
-namespace lth_locale = leatherman::locale;
 
 // Declare environ for OSX
 extern char** environ;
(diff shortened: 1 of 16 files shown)


execution/src/posix/execution.cc
```diff
@@ -279,7 +282,7 @@ namespace leatherman { namespace execution {
 
         // Set the process group; this will be used by the parent if we need to kill the process and its children
         if (setpgid(0, 0) == -1) {
-            char const* message = "failed to setpgid.";
+            char const* message = _("failed to setpgid.").c_str();
```
MikaelSmith: I'm not sure this is safe; alse we could use `size` instead of running `strlen`. I'd rather we not try to hold onto a pointer from `c_str`.

json_container/src/json_container.cc
```diff
@@ -301,12 +305,12 @@ namespace leatherman { namespace json_container {
...
                                                     const char* key) const {
         if (!jval.IsObject()) {
-            throw data_type_error { "not an object" };
+            throw data_type_error { _("not an object") };
         }
 
         if (!jval.HasMember(key)) {
-            std::string err_msg { "unknown object entry with key: " };
-            throw data_key_error { err_msg + key };
+            std::string err_msg { _("unknown object entry with key: {1}", key) };
```
MikaelSmith: The intermediate object is no longer needed.

logging/src/logging.cc
```diff
@@ -198,7 +201,7 @@ namespace leatherman { namespace logging {
                 return in;
             }
         }
-        throw runtime_error(lth_locale::format("invalid log level '%1%': expected none, trace, debug, info, warn, error, or fatal.", value));
```
MikaelSmith: Good catch.

execution/src/posix/execution.cc
ahenroid: You're right, this isn't safe. I'll clean up the few instances of this.

windows/src/wmi.cc
```diff
@@ -25,9 +27,9 @@ namespace leatherman { namespace windows {
     {
 #ifdef LEATHERMAN_I18N
         // LOCALE: format a pointer as hex for printing an error message.
-        return lth_locale::format("{1} (0x{2,num=hex})", s, hres);
+        return _("{1} (0x{2,num=hex})", s, hres);
 #else
-        return lth_locale::format("%1% (0x%2$#x)", s, hres);
+        return _("%1% (0x%2$#x)", s, hres);
```
branan: I think lth_locale does the right thing here now, and we shouldn't need to special-case format strings anymore

MikaelSmith: boost::format and boost::locale::format have different special formats, that's why it does `%2$#x` and `{2,num=hex}`. I'm not aware of a way to do this in a compatible way, and handling all the special cases to convert would be crazy. So I think this special case is still necessary.

ahenroid: Yes, I just looked at this again and agree that these lines have to remain like this. Our fix-up code in leatherman::locale::format for non-LEATHERMAN_I18N platforms does not handle the formatting here. I believe this is the one and only place we need this.

I'd say merge this PR unless there are any other concerns.

PR #203, opened 2016-08-23
(LTH-112)  Add thread and schema validation utilities

MikaelSmith:


MikaelSmith:
This might be a simpler PR after https://github.com/puppetlabs/leatherman/pull/204.

MikaelSmith:
Coverage on a rapidjson/valijson adapter class is low. That was eventually upstreamed, so we could remove that code by bumping to a newer valijson version.

PR #204, opened 2016-08-26
(maint) Switch to using compressed vendored packages

MikaelSmith:
Switches to using compressed vendored packages to make git history less
noisy when changing versions. Also simplify CMake files for vendored
libraries with a new macro.

Split into two commits so they're easier to review.


MikaelSmith:
This may have a side-effect of giving us more honest coverage numbers. I think rapidjson, nowide, and catch were probably pretty well covered by unit tests.

MikaelSmith:
Oh right, compiling a library is a problem with this pattern. Only works for header-only libraries.

MikaelSmith:
Probably needs some work to get dependencies right.

MikaelSmith:
@branan thoughts on this idea?

ahenroid:
This sounds like a great idea. Is there no way to apply this also to compiled libraries?

MikaelSmith:
@ahenroid extracting the packages happens during a build, not CMake configuration, and CMake doesn't make a special case for dependencies. CMake will fail to identify the right build order and in fact will error because the source file we want to build isn't available yet.

MikaelSmith:
I'm not sure coveralls data is being uploaded correctly :( looking into it.

MikaelSmith:
Fixed code coverage flags.

MikaelSmith:
This could switch to using git-lfs...

kylog:
After hipchatter everyone was thumbs-up on this.

PR #205, opened 2016-08-31
(LTH-113) Fix building Leatherman without cflags

MikaelSmith:
The cflags CMake module includes useful settings, but shouldn't be
required to build against Leatherman. Fix compilation so that's true.


Magisus:
Seems good, I'll merge this and hope it fixes the user's issues. Thanks!

MikaelSmith:
They can pull from source until we do a release.

PR #208, opened 2016-09-16
(LTH-114) Allow stripping of CR characters on Windows

branan:


execution/inc/leatherman/execution/execution.hpp
```diff
@@ -430,6 +434,7 @@ namespace leatherman { namespace execution {
      */
     std::tuple<std::string, std::string> process_streams(
             bool trim,
+            bool strip_cr,
```
branan: Technically, this is an API change, but I can't imagine that anyone actually uses `process_streams` directly.

I'd actually consider moving this into an internal header "soon" to be a pretty good plan, since it's an implementation detail of the platform-specific execution codepaths

branan: @MikaelSmith Do you have any thoughts or additional context on this? This feels like a "safe" change to me given how specific to the guts of execution this function actually is.

branan:
Besides my semver question above, this has been tested in a pooler machine and appears to work for facter#1424

glennsarti:
Should there be any unit tests around this behaviour? Seems like there should.

execution/inc/leatherman/execution/execution.hpp
MikaelSmith: Yeah, I think we should call it out as a change but not worry about planning for the change (as no one should be impacted by it). We could throw it in an anonymous namespace to make it hard to use.

MikaelSmith: On a related note, https://github.com/puppetlabs/facter/blob/master/lib/inc/internal/execution/execution.hpp looks like it should've been removed (it's not used anywhere). I'll throw up a PR.

Magisus:
@branan re: our conversation in standup this morning, I personally like having this handled within Leatherman. I don't think it's too magical; I think it's finicky detail work I'd rather not have to deal with in my main code, hence the whole purpose of using a utility library.

But I agree, tests would be nice, so that at least there's some "documentation" around this behavior.

execution/tests/windows/execution.cc
```diff
@@ -375,6 +375,23 @@ SCENARIO("executing commands with execution::execute") {
...
+    GIVEN("a command that outputs windows-style newlines") {
+        // These are the default options so that I don't override them.
+        lth_util::option_set<execution_options> options = { execution_options::trim_output, execution_options::merge_environment, execution_options::redirect_stderr_to_null };
+        WHEN("newlines are not normalized") {
+            auto exec = execute("cmd.exe", { "/c", "type", normalize(EXEC_TESTS_DIRECTORY "/fixtures/ls/fcrlf.txt") }, 0, options);
+            REQUIRE(exec.success);
+            REQUIRE(exec.error == "");
+            REQUIRE(exec.output.find('\r') != std::string::npos);
+        }
+        WHEN("requestecd to normalize newlines") {
```
Magisus: Typo here

branan:
@Magisus @MikaelSmith tests added. These have passed appveyor once, and should again - I just force-pushed for the typo fix.

Magisus:
👍  from me once the checks are green

execution/src/execution.cc
```diff
@@ -350,13 +350,22 @@ namespace leatherman { namespace execution {
...
-    static bool process_data(bool trim, string const& data, string& buffer, string const& logger, function<bool(string&)> const& callback)
+    static bool process_data(bool trim, bool strip_cr, string const& raw_data, string& buffer, string const& logger, function<bool(string&)> const& callback)
     {
         // Do nothing if nothing was read
-        if (data.empty()) {
+        if (raw_data.empty()) {
             return true;
         }
 
+        string stripped_data;
```
branan: I believe that if I made this static, we would avoid having to allocate every loop (at some point it would just end up at the output buffer size, and continue to be overwritten in that location). The tradeoff there would be that we'd leak that 4K, as there's no object to tie the lifetime of this variable to.

A third option is to make the series of `data` arguments eventually leading to this function non-const. That would allow us to modify the buffer in place.

Yet a fourth would be to try moving this logic to the `rw_from_child` method in windows.cc, which may actually be the best option - it keeps this windows-focused code in the right place, and avoids the need to change the "public" API mentioned above

branan: I think I have convinced myself to move this logic to `rw_from_child`, actually.

execution/src/windows/execution.cc
```diff
@@ -313,6 +313,11 @@ namespace leatherman { namespace execution {
                     if (pipe.read) {
                         // Read completed immediately, process the data
                         pipe.buffer.resize(count);
+                        if (convert_newlines) {
+                            pipe.buffer.erase(
+                                std::remove(pipe.buffer.begin(), pipe.buffer.end(), '\r'),
```
MikaelSmith: Are there ever times you'd legitimately want the CR (when clearing end-of-line CR)? I can't think of any.

branan: I contemplated this. I guess it's possible that some program meant for Unix might spit out CRs in places, but have its LFs magically transformed to CRLF pairs. I can defnitely change this to the proper "only convert pairs" logic, it's just a bit more compllicated than a nice algorithm application

Magisus:
The refactor appears to have broken the tests.

branan:
@Magisus I'm aware, and very very confused by it. I'm working on tracking it down

MikaelSmith:
Did you see my comment about line 391? Might be a last-line problem.

branan:
@Magisus This is green now! Looks like I didn't fully understand the windows codepaths - there's some weirdness/duplication there when IO is "pending" that puts us into an async mode (which I don't fully understand).

Just duplicating my cr conversion to both places where we can have new data come in works, but I'm not sure I like the code duplication in that function - I'd really like somebody who knows windows I/O to talk me through it

MikaelSmith:
@branan I wrote it, but not sure it's ideal. It's dealing with reading from multiple pipes synchronously. Maybe someone else can suggest a better way to approach it.

branan:
@MikaelSmith Hah, your comment finally came through. That would've saved me some effort yesterday. Oh well, at least I have a decent windows dev environment now (more or less)

Magisus:
👍

PR #212, opened 2016-10-24
(LTH-117) Switch to compile-time unpacking vendored packages

MikaelSmith:
CMake doesn't handle include file dependencies with ExternalProject_Add.
Switch to unpacking files at compile-time so they can be used in
dependency resolution.


prometheanfire:
This fix looks like it fixed the build for me.

Magisus:
And likewise, can confirm it still builds for me. Seems reasonable 👍

PR #213, opened 2016-10-25
(LTH-118) Remove Leatherman removing the lib prefix off targets 

jmslagle:
it installs.  This makes it generate libraries in expected places on unix line OSs.  NOTE:  This has downstream project implications that need managed - I am working on CPP-HOCON and a patch for that should be forthcoming


paravoid:
Debian packaging would require this as well.

GabrielNagy:
Closing this too, see https://github.com/puppetlabs/cpp-hocon/pull/86 for reference

PR #216, opened 2016-11-14
(maint) explicitly pass release flag to pod2man

wkalt:
This is a followup on
https://github.com/puppetlabs/orchestrator-client/pull/270

This flag is required when using podman >= 4 or perl 5.24, although I
don't think it has any consequence in leatherman right now since there
are no man pages.


nicklewis:
:+1:

PR #217, opened 2016-12-01
(LTH-120) Generalized detached process

mruzicka:
This PR renames the `create_new_process_group` execution option to `create_detached_process`.
The intention of the option was to create a process in such a way it is detached from its parent in terms of their lifespan. On Windows, which was the only platform where the option was implemented, the child process would be killed when the parent process exited in case the option was not specified. On the contrary the child can outlive its parent when the option is specified.
The renaming of the option is a preparation for generalization of the concept to other OSs where it is also applicable (notably Solaris) but where the original name would not make sense or would be misleading.
The other commit in this PR provides the Solaris implementation of the option which mainly entails creating separate process contracts for the child processes.

### DIFF
--- a/execution/CMakeLists.txt
+++ b/execution/CMakeLists.txt
@@ -1,6 +1,12 @@
 find_package(Boost 1.54 REQUIRED COMPONENTS regex filesystem system)
 
 add_leatherman_deps("${Boost_LIBRARIES}")
+if ("${CMAKE_SYSTEM_NAME}" MATCHES "SunOS")
+    # We use functions provided by this library in the implementation
+    # of the create_detached_process execution option on Solaris to
+    # execute the child processes in their own contracts
+    add_leatherman_deps(contract)
+endif()
 add_leatherman_includes("${Boost_INCLUDE_DIRS}")
 
 leatherman_dependency(util)
@@ -21,13 +27,20 @@ add_leatherman_headers(inc/leatherman)
 if(WIN32)
     add_leatherman_library(src/execution.cc src/windows/execution.cc)
 else()
-    add_leatherman_library(src/execution.cc src/posix/execution.cc)
+    if("${CMAKE_SYSTEM_NAME}" MATCHES "SunOS")
+        add_leatherman_library(src/execution.cc src/posix/execution.cc src/posix/solaris/platform.cc)
+    else()
+        add_leatherman_library(src/execution.cc src/posix/execution.cc src/posix/generic/platform.cc)
+    endif()
 endif()
 
 if(WIN32)
     set(PLATFORM_TESTS tests/windows/execution.cc)
 else()
     set(PLATFORM_TESTS tests/posix/execution.cc)
+    if("${CMAKE_SYSTEM_NAME}" MATCHES "SunOS")
+        list(APPEND PLATFORM_TESTS tests/posix/solaris/execution.cc)
+    endif()
 endif()
 
 add_leatherman_test(tests/log_capture.cc ${PLATFORM_TESTS})
--- a/execution/inc/leatherman/execution/execution.hpp
+++ b/execution/inc/leatherman/execution/execution.hpp
@@ -57,9 +57,10 @@ namespace leatherman { namespace execution {
          */
         preserve_arguments = (1 << 7),
         /**
-         * On Windows, create a new process group for the child process, but not a Job Object.
+         * Create a new process such that it can outlive its parent. This involves running it
+         * in a separate process group on Windows, and in a separate contract on Solaris.
          */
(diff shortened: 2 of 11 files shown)


execution/src/posix/solaris/platform.cc
```diff
@@ -0,0 +1,197 @@
...
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
```
mruzicka: I wonder if this header is at all necessary, given how much drifted from the file it was based on: https://github.com/openssh/openssh-portable/blob/V_7_3/openbsd-compat/port-solaris.c

execution/CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 find_package(Boost 1.54 REQUIRED COMPONENTS regex filesystem system)
 
 add_leatherman_deps("${Boost_LIBRARIES}")
+if ("${CMAKE_SYSTEM_NAME}" MATCHES "SunOS")
+    add_leatherman_deps(contract)
```
MikaelSmith: Please add a comment about why this is needed.

execution/inc/leatherman/execution/execution.hpp
```diff
@@ -57,9 +57,10 @@ namespace leatherman { namespace execution {
          */
         preserve_arguments = (1 << 7),
         /**
-         * On Windows, create a new process group for the child process, but not a Job Object.
+         * Create a new process such that it can outlive its parent. This involves running it
+         * in a separate process group on Windows, and in a sparate contract on Solaris.
          */
-        create_new_process_group = (1 << 8),
+        create_detached_process = (1 << 8),
```
MikaelSmith: This is a breaking change, but one that's unused. I'm comfortable including this in 0.10.0 or 1.0.0.

execution/src/posix/solaris/platform.cc
MikaelSmith: I feel more comfortable leaving it.

execution/src/posix/solaris/platform.cc
```diff
@@ -0,0 +1,197 @@
...
+
+        return tmpl_fd;
+
+    close_fail:
+        close(tmpl_fd);
+    fail:
+        throw execution_exception(format_error(_("failed to create process contract template"), err));
+    }
+
+    static int deactivate_contract_tempalte(int tmpl_fd)
```
MikaelSmith: Typo?

execution/CMakeLists.txt
mruzicka: ok

execution/src/posix/solaris/platform.cc
mruzicka: oops

execution/src/posix/solaris/platform.cc
mruzicka: ok

mruzicka:
I've addressed the comments.

MikaelSmith:
Have you checked this works on Solaris 11?

execution/inc/leatherman/execution/execution.hpp
```diff
@@ -57,9 +57,10 @@ namespace leatherman { namespace execution {
          */
         preserve_arguments = (1 << 7),
         /**
-         * On Windows, create a new process group for the child process, but not a Job Object.
+         * Create a new process such that it can outlive its parent. This involves running it
+         * in a separate process group on Windows, and in a sparate contract on Solaris.
```
MikaelSmith: Typo in "sparate".

mruzicka:
I developed it on Solaris 11 and tested in on Solaris 10 yesterday.

mruzicka:
I've fixed the typo and added changes of translation files, which only got updated today for some reason.

PR #222, opened 2016-12-06
(LTH-120) Isolate lfs to Leatherman.execution

MikaelSmith:
Solaris large file support is needed for the ctrun functionality in
Leatherman.execution. It's tricky to apply everywhere however, which
caused building Facter with this version of Leatherman to fail. Restrict
large file support flags to Leatherman.execution.

Only applying lfs to some files should be safe according to
http://docs.oracle.com/cd/E19455-01/806-0634/6j9vo5alu/index.html

```
Applications may combine objects produced in the large file compilation
environment with objects produced in the transitional compilation
environment, but must be careful with respect to interoperability
between those objects. Applications should not declare global variables
of types whose sizes change between compilation environments.
```

### DIFF
--- a/cmake/cflags.cmake
+++ b/cmake/cflags.cmake
@@ -42,11 +42,7 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
 
     # missing-field-initializers is disabled because GCC can't make up their mind how to treat C++11 initializers
     set(LEATHERMAN_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Werror -Wno-unused-parameter -Wno-unused-local-typedefs -Wno-unknown-pragmas -Wno-missing-field-initializers")
-    if ("${CMAKE_SYSTEM_NAME}" MATCHES "SunOS")
-        # this is needed to link against the libcontract library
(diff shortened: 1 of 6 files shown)


MikaelSmith:
Passed a full Solaris 11 build @mruzicka

cmake/cflags.cmake
```diff
@@ -42,11 +42,7 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
 
     # missing-field-initializers is disabled because GCC can't make up their mind how to treat C++11 initializers
     set(LEATHERMAN_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Werror -Wno-unused-parameter -Wno-unused-local-typedefs -Wno-unknown-pragmas -Wno-missing-field-initializers")
-    if ("${CMAKE_SYSTEM_NAME}" MATCHES "SunOS")
-        # this is needed to link against the libcontract library
```
mruzicka: I suggest expanding the comment to something like:
```
        # this is needed to compile the posix/solaris/platform.cc such that
        # it links correctly against the libcontract library
```

MikaelSmith: Comment expanded below.

MikaelSmith:
Test build using this started at https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_pkg-van-ship_manual-ad-hoc/432/, based on https://github.com/MikaelSmith/puppet-agent/tree/test-pxp.

PR #223, opened 2016-12-09
(PCP-657) Merge master to LTS-1.7

mruzicka:


mruzicka:
Since this is a fast-froward, let's avoid the extra commit and push directly.

PR #224, opened 2016-12-09
(maint) Avoid heap allocations during exec

MikaelSmith:
Removes string localization in forked child exec, as heap allocation is
not safe there.

### DIFF
--- a/execution/src/posix/execution.cc
+++ b/execution/src/posix/execution.cc
@@ -281,11 +281,12 @@ namespace leatherman { namespace execution {
         // Do not allocate heap memory or throw exceptions
         // The child is sharing the address space of the parent process, so carelessly modifying this
         // function may lead to parent state corruption, memory leaks, and/or total protonic reversal
+        // As such, strings are explicitly not localized in this function.
 
         // Set the process group; this will be used by the parent if we need to kill the process and its children
         if (setpgid(0, 0) == -1) {
-            string message = _("failed to setpgid.");
-            if (write(err_fd, message.c_str(), message.size()) == -1) {
+            char const* message = "failed to setpgid.";
+            if (write(err_fd, message, strlen(message)) == -1) {
                 // Do not care
             }
             return;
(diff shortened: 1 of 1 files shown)


mruzicka:
Is this causing problems? I thought that temporary allocations - i.e. allocations which are freed before the `exec`, `_exit` calls - could actually be OK.
If not than similar change needs to be done here: https://github.com/puppetlabs/leatherman/blob/master/execution/src/posix/solaris/platform.cc#L175

MikaelSmith:
Ok, would need to update that as well.

These are @peterhuene's observation. Is there a good source for the restriction?

mruzicka:
I don't know about any.

peterhuene:
Provided a thread-safe libc is being used (and it is in our case), there is no inherent problem in allocating memory if it is freed prior to the `_exit` call (note that specific call is intentional; we don't want to invoke parent `atexit`s from the child process upon failure), the intention of this code was to be as simplistic as possible as to not introduce memory leaks or corruption in the parent, which, frankly, are easily done in vfork'd code.

However, the default allocator in C++ is a throwing allocator.  Can we guarantee the stdlib's `std::terminate` and `std::abort` will not corrupt parent process state if we fail to allocate memory for one of these localized error messages?  Personally I prefer to limit what we do from the vfork'd child as to not provide a basis for future complex operations vs. localizing error messages that should never be displayed.

peterhuene:
To wit, there weren't any problematic observations and I consider the existing code to be correct.  Ultimately these rules were designed for future maintainers of this code so that they don't introduce subtle and hard to diagnose bugs into the parent process.  As stated, I'd personally like to just adhere to these rules for that reason alone, even if it means that we can't localize a few strings, but I can see the value in having this be correct in terms of localization too even if it means we relax the rules.

peterhuene:
Sorry for the numerous edits to the previous comments 😞; it takes me some time before they stop reading like _Ulysses_.

MikaelSmith:
We already have some messages that aren't externalized because they're printed when locale setup fails. I'm comfortable having a few more, but I'd also be fine with commenting these well to be clear why they're safe.

PR #226, opened 2017-01-26
(LTH-125) Avoid dynamic symbols between vfork/exec

MikaelSmith:


execution/src/posix/execution.cc
```diff
@@ -463,11 +448,26 @@ namespace leatherman { namespace execution {
                                             options[execution_options::inherit_locale]);
         auto envp = to_exec_arg(&variables);
 
+        // Set all open file descriptors above stderr to close on exec
+        auto max_desc_limit = get_max_descriptor_limit();
```
MikaelSmith: @peterhuene does this seem like a sane replacement for closing the file descriptors in the child?

peterhuene: Isn't this a timing issue where one thread marked the descriptors close-on-exec just prior to the `vfork`, another thread just completed a `vfork` and is now marking the same descriptors as not close-on-exec just in time for the first thread to `exec` without them flagged? 

Solaris is the worst, but Linux stubbornly not providing a `closefrom` equivalent is equally maddening (no really, we only want three open descriptors for our child process!).

execution/src/posix/execution.cc
```diff
@@ -281,48 +281,33 @@ namespace leatherman { namespace execution {
...
+        // As such, strings are explicitly not localized in this function.
+        //
+        // This is especially important due to a deadlock in vfork/exec on Solaris, identified in
+        // http://www.oracle.com/technetwork/server-storage/solaris10/subprocess-136439.html. The solution
+        // they use in posix_spawn is to avoid calling functions exported as from libc as global symbols
+        // after the fork. `write`, `strlen`, and `close` are still suspect below.
 
         // Set the process group; this will be used by the parent if we need to kill the process and its children
         if (setpgid(0, 0) == -1) {
-            string message = _("failed to setpgid.");
```
MikaelSmith: We lose some logging, but we still have errno (which all of these calls set) in the exit code. I think there's minimal extra value from the messages.

execution/src/posix/execution.cc
MikaelSmith: Yeah. I don't think we exec in parallel anywhere, but I could add a (global) mutex to guard it. I'm pretty sure we shouldn't do being this stuff in parallel anyway.

MikaelSmith: Updated with a mutex.

MikaelSmith:
@puppetlabs/leatherman-maintainers I'd like to go forward with this as the fix for PCP-695. We may need further work (I can't guarantee this will never deadlock on Solaris), but right now it's causing pain we should fix right away.

Magisus:
It looks sane enough. Since the ad hoc run went green, it certainly seems to give us a better shot than what we have right now. I don't really have the details to judge the solution, but I'm 👍  to getting it through in the interest of the release.

MikaelSmith:
@peterhuene gave a guarded thumbs-up in HipChat.

PR #227, opened 2017-01-31
(LTH-126) Make it possible to use fork() on demand

mruzicka:
Traditionally leatherman::execution used `vfork()` for the child process creation because of its higher performance and lower virtual memory requirements. But this turned out to be problematic (deadlock prone) in multi-threaded processes on at least Solaris.
So in this PR a new execution option - `thread_safe` - is introduced, which when specified causes leatherman::execution to use slower but safer `fork()` for the child process creation. If the option is not specified the more performant but more dangerous `vfork()` is used to preserve backward compatibility.

### DIFF
--- a/execution/CMakeLists.txt
+++ b/execution/CMakeLists.txt
@@ -1,4 +1,4 @@
-find_package(Boost 1.54 REQUIRED COMPONENTS thread regex filesystem system)
+find_package(Boost 1.54 REQUIRED COMPONENTS regex filesystem system)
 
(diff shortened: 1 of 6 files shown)


MikaelSmith:
That seems straight-forward. You decided not to change back how we release file descriptors?

Also, how have you tested that this solves the original problem?

mruzicka:
@MikaelSmith I'm going to add the revert now and I'll create a new pxp-agent build tomorrow and work with James to verify we don't get the deadlock any more.

PR #229, opened 2017-02-22
(LTH-128) Use less Boost.Log magic

MikaelSmith:
On AIX, when used with pxp-agent, logging doesn't work because the
log-level (severity) isn't correctly stored by the severity logger. This
is strange as it does work with Facter; I suspect the cause is due to
differences in linking with a .exe (pxp-agent) vs a .so (libfacter)
against libboost_logging.so. Work around the issue in Boost.Log by using
less of its magic, and instead using an explicit attribute for severity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.11.1]
+
+### Fixed
+- Circumvent a bug in Boost.Log's severity logger that prevented logging on AIX (LTH-128)
+
 ## [0.11.0]
 
 ### Added
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.2.2)
-project(leatherman VERSION 0.11.0)
+project(leatherman VERSION 0.11.1)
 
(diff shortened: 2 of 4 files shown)


MikaelSmith:
I plan to follow up with a build on AIX using this tomorrow to re-confirm that it fixes PCP-677.

MikaelSmith:
/code
[0] [AIX] root@pe-aix-71-client:/var/tmp/tmp.i6RPLgw4en # /opt/puppetlabs/puppet/bin/pxp-agent --foreground --logfile - --loglevel trace
2017-02-22 11:37:56.582444 DEBUG puppetlabs.pxp_agent.configuration:271 - Logging configured
2017-02-22 11:37:56.584192 DEBUG puppetlabs.pxp_agent.main:186 - pxp-agent logging has been initialized
2017-02-22 11:37:56.593999 ERROR puppetlabs.pxp_agent.main:205 - Fatal configuration error: broker-ws-uri or broker-ws-uris must be defined; cannot start pxp-agent

MikaelSmith:
ping @puppetlabs/leatherman-maintainers

ahenroid:
+1 from me

Magisus:
👍

Magisus:
Which branch of puppet-agent do we want this version bump to go into?

MikaelSmith:
This fixes PCP-677, which applies to all releases. I'd opt for getting it in everywhere. That would require some extra work for LTS, so I'd be fine with stable/master for now.

MikaelSmith:
Specifically, using Leatherman 0.11 in the LTS branch would require cherry-picking PCP-704.

PR #230, opened 2017-02-24
Use GNUInstallDirs to install to /usr/lib/$triplet

paravoid:
Debian-based distributions (including Ubuntu) nowadays have their
library directory under /usr/lib/<triplet>, e.g.
/usr/lib/x86_64-linux-gnu/.  /usr/lib64 doesn't exist in Debian.

Use CMake's GNUInstallDirs and CMAKE_INSTALL_LIBDIR to find the libdir
instead of hardcoding lib${LIB_SUFFIX}.

This most likely breaks Windows builds -- and possibly other Linux
builds too; I only have Debian systems in my possession. I hope it's at
least a start for someone more knowledgeable with both other platforms
and CMake to fix :)

### DIFF
--- a/cmake/leatherman.cmake
+++ b/cmake/leatherman.cmake
@@ -1,6 +1,8 @@
 # This file contains utilities used by both leatherman and consuming
 # projects.
 
+include(GNUInstallDirs)
+
 # Save the directory this script is from, to reference other files
 # located in the same directory when using cmake in script mode.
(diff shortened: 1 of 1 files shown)


paravoid:
I think this contribution is trivial :)

MikaelSmith:
Sounds like a good idea. I think we can evaluate this alongside https://github.com/puppetlabs/leatherman/pull/213.

GabrielNagy:
Closing this, see https://github.com/puppetlabs/cpp-hocon/pull/86 for reference

PR #231, opened 2017-02-24
Remove -Werror

paravoid:
-Wall -Wextra -Werror is a bit extreme. Compilers frequently add new
warning messages between versions and having the build fail every time
this happens is problematic.

I patched this for Debian but I now see that the FreeBSD proposed port
has also done the very same, so I suppose the sentiments are shared.


paravoid:
I think this contribution is trivial :)

MikaelSmith:
We'd really like to keep this as the default. Does something like `CMAKE_CXX_FLAGS='-Wno-error'` work to avoid the issue?

MikaelSmith:
For now, an alternative would be to specifically disable the warnings you're running into in `CMAKE_CXX_FLAGS`.

MikaelSmith:
Filed [LTH-131](https://tickets.puppetlabs.com/browse/LTH-131) to allow overriding all compile flags.

PR #232, opened 2017-03-06
(LTH-131) Allow overriding CMAKE_CXX_FLAGS

MikaelSmith:
Prepends all Leatherman config to CMAKE_CXX_FLAGS, so that the
command-line can override any of the options we specify.


Magisus:
Just to make sure I understand this, `CMAKE_CXX_FLAGS` contains things coming in from the user's compile command?

MikaelSmith:
Yes. Someone wanted to override using `-Werror` everywhere to support newer compilers. They could do that by setting `CMAKE_CXX_FLAGS=-Wno-error`, except that our `-Werror` comes after and overrides it.

PR #233, opened 2017-03-29
(LTH-133) Allow using an environment variable to lookup locale files

MikaelSmith:
Allows configuring an environment variable to use at runtime to find
locale files. This environment variable will usually be set by a script
wrapping invocation on Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,21 +41,28 @@ endif()
 defoption(LEATHERMAN_MOCK_CURL "Use mock curl library for testing Leatherman.curl" ${MOCK_CURL})
 
 add_definitions(${LEATHERMAN_DEFINITIONS})
-add_definitions(-DLEATHERMAN_LOCALE_INSTALL="${CMAKE_INSTALL_PREFIX}/share/locale")
+if (LEATHERMAN_LOCALE_VAR AND LEATHERMAN_LOCALE_INSTALL)
+    # Add an environment variable to look up install prefix at runtime.
+    add_definitions(-DLEATHERMAN_LOCALE_VAR="${LEATHERMAN_LOCALE_VAR}"
+                    -DLEATHERMAN_LOCALE_INSTALL="${LEATHERMAN_LOCALE_INSTALL}")
+else()
+    # Add install location instead.
+    add_definitions(-DLEATHERMAN_LOCALE_INSTALL="${CMAKE_INSTALL_PREFIX}/share/locale")
+endif()
 
 file(GLOB_RECURSE ALL_LEATHERMAN_SOURCES */src/*.cc */inc/*.hpp)
 add_subdirectory(locales)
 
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -41,21 +41,28 @@ endif()
...
 
 add_definitions(${LEATHERMAN_DEFINITIONS})
-add_definitions(-DLEATHERMAN_LOCALE_INSTALL="${CMAKE_INSTALL_PREFIX}/share/locale")
+if (LEATHERMAN_LOCALE_VAR AND LEATHERMAN_LOCALE_INSTALL)
+    # Add an environment variable to look up install prefix at runtime.
+    add_definitions(-DLEATHERMAN_LOCALE_VAR="${LEATHERMAN_LOCALE_VAR}"
+                    -DLEATHERMAN_LOCALE_INSTALL="${LEATHERMAN_LOCALE_INSTALL}")
+else()
+    # Add install location instead.
+    add_definitions(-DLEATHERMAN_LOCALE_INSTALL="${CMAKE_INSTALL_PREFIX}/share/locale")
```
Magisus: In the general case in a vanagon project, what is CMAKE_INSTALL_PREFIX here?

Magisus:
Can this only be specified when leatherman is built, or can it be overridden for each project that brings in leatherman as a library? (Still trying to work out how CMake strings everything together. Also not sure if that question will even be relevant.)

CMakeLists.txt
MikaelSmith: https://cmake.org/cmake/help/v3.0/variable/CMAKE_INSTALL_PREFIX.html states it defaults to `/usr/local` or `C:/Program Files`.

MikaelSmith:
Only when Leatherman is built.

CMakeLists.txt
Magisus: So if we don't set these environment vars on Unix (was looking at the linked client-tools PR), it will look in `/usr/local/share/locale` for translation files? I'm pretty sure at least for puppet-agent that is not what we want, so I assume we will need to set these for Leatherman there as well, probably on all platforms, to the paths stated [here](https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md)?

MikaelSmith: https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/leatherman.rb#L88

MikaelSmith: If you don't specify LEATHERMAN_LOCALE_INSTALL and LEATHERMAN_LOCALE_VAR when configuring Leatherman, it'll use CMAKE_INSTALL_PREFIX to generate a fixed location. We already set that on non-Windows platforms. We set it on Windows too, but Windows packages are relocatable so we don't rely on it.

Magisus: Okay, got it, thanks!

MikaelSmith:
Added release prep for 0.11.2.

Magisus:
👍

PR #234, opened 2017-04-10
(LTH-134) Add rb_ll2inum binding

branan:
rb_int2inum usually takes a `long`, which can be 32-bits on some
platforms. rb_ll2inum will always give us a 64-bit value, if it's
available.

### DIFF
--- a/ruby/inc/leatherman/ruby/api.hpp
+++ b/ruby/inc/leatherman/ruby/api.hpp
@@ -266,6 +266,10 @@ namespace leatherman {  namespace ruby {
         /**
          * See MRI documentation.
          */
+        VALUE (* const rb_ll2inum)(LONG_LONG);
+        /**
+         * See MRI documentation.
+         */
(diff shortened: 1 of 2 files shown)


ruby/inc/leatherman/ruby/api.hpp
```diff
@@ -266,6 +266,10 @@ namespace leatherman {  namespace ruby {
         /**
          * See MRI documentation.
          */
+        VALUE (* const rb_ll2inum)(SIGNED_VALUE);
```
MikaelSmith: SIGNED_VALUE seems incorrect, since that's intptr_t (32-bit on 32-bit systems).

branan: oh, whoops. Copy-paste error. let me check what the actual signature of that function is

branan:
@MikaelSmith Try it now

PR #237, opened 2017-04-12
(LTH-135) Make gettext_compile respect LOCALE env vars

Magisus:
This commit fixes an issue where if the user set LEATHERMAN_LOCALE_VAR
and LEATHERMAN_LOCALE_INSTALL to determine where leatherman should look
for translation files, the install of those files from consuming
projects would not respect those vars and install them in the default
location, which may not correspond to the dir specified with those vars.
Now if the vars are set, leatherman will both install to the specificed
directory and read translations from it at runtime.

### DIFF
--- a/cmake/leatherman.cmake
+++ b/cmake/leatherman.cmake
@@ -239,6 +239,10 @@ endmacro()
 # Compile gettext .po files into .mo files and configure installing to inst
 # Creates a custom target `translations`.
 #
+# If LEATHERMAN_LOCALE_INSTALL and LEATHERMAN_LOCALE_VAR are set, this will
+# install the newly compiled .mo files to the location specified by these vars.
+# Otherwise, it will use the specified `inst` arg.
+#
 # Does nothing if msgfmt (part of gettext) isn't found. Sets GETTEXT_ENABLED
 # to ON if we can compile .mo files, otherwise sets to OFF. This variable can
 # be used to disable functionality (such as testing) that requires gettext
@@ -273,7 +277,11 @@ macro(gettext_compile dir inst)
                 DEPENDS ${fpath})
             add_custom_target(${lang}-${PROJECT_NAME} DEPENDS ${mo})
             add_dependencies(translations ${lang}-${PROJECT_NAME})
-            install(FILES ${mo} DESTINATION "${inst}/${lang}/LC_MESSAGES")
+            if(LEATHERMAN_LOCALE_VAR AND LEATHERMAN_LOCALE_INSTALL)
+                install(FILES ${mo} DESTINATION "${LEATHERMAN_LOCALE_VAR}/${LEATHERMAN_LOCALE_INSTALL}/${lang}/LC_MESSAGES")
+            else()
+                install(FILES ${mo} DESTINATION "${inst}/${lang}/LC_MESSAGES")
+            endif()
         endforeach()
         set(GETTEXT_ENABLED ON)
     else()


cmake/leatherman.cmake
```diff
@@ -273,7 +277,11 @@ macro(gettext_compile dir inst)
                 DEPENDS ${fpath})
             add_custom_target(${lang}-${PROJECT_NAME} DEPENDS ${mo})
             add_dependencies(translations ${lang}-${PROJECT_NAME})
-            install(FILES ${mo} DESTINATION "${inst}/${lang}/LC_MESSAGES")
+            if(LEATHERMAN_LOCALE_VAR AND LEATHERMAN_LOCALE_INSTALL)
+                install(FILES ${mo} DESTINATION "${LEATHERMAN_LOCALE_VAR}/${LEATHERMAN_LOCALE_INSTALL}/${lang}/LC_MESSAGES")
```
MikaelSmith: I don't think this does what you think it does...

Magisus: Interesting, it worked in my test...

MikaelSmith: `LEATHERMAN_LOCALE_VAR` is expected to contain the name of an enviroment variable, that Leatherman reads at runtime to determine a starting path (with `LEATHERMAN_LOCALE_INSTALL` used to define a project-determined offset from that location).

MikaelSmith: I think we need to take LEATHERMAN_LOCALE_INSTALL into account somehow, but either need to use CMAKE_INSTALL_PREFIX or a new variable to determine where to put the files for packaging.

Magisus: Ah okay I see. So I was using the var wrong.

MikaelSmith: It would pass unit tests, but not work in a fresh install.

Magisus: So then it seems like coupling these two may actually be impossible. Which means we may have to fall back on platform-specific install instructions in consuming projects. Ah well.

Magisus: CMAKE_INSTALL_PREFIX might work.

Magisus:
So, CMake's `install` function already prepends CMAKE_INSTALL_PREFIX when passed a relative path. The problem we're running into in puppet-agent is that Facter's prefix is different from Leatherman's on Windows. While we could solve this by changing the `compile_gettext` call in Facter to `../puppet/share/locale` on Windows, that feels very wrong, seeing as it makes no sense in any context besides the agent. So I think I'm going to go the route of adding a new CMake var that Facter will respect, to allow maximum flexibility.

MikaelSmith:
The other piece that needs to stay insync is where Leatherman is configured to look for files. Having multiple places (determined per project) is complicated (though not impossible, that's what https://github.com/puppetlabs/leatherman/blob/master/locale/inc/leatherman/locale/locale.hpp#L47 does to support unit testing).

It makes sense in my mind for Leatherman to define one place to install and look for locale files, used by all projects that make use of Leatherman.

Magisus:
That makes sense to me too, do you have an idea how to coordinate that? Leatherman is built separately from any project that consumes it, and the two don't have any insight into each other's configuration. So what are our other options?

MikaelSmith:
https://github.com/Magisus/leatherman/blob/05751046cfc49eaec8a3b2121f2cc4194e3d9e37/cmake/leatherman.cmake#L6 appears to preserve a variable for later use. We could do the same thing with CMAKE_INSTALL_PREFIX when installing Leatherman, and use that in the macro.

Magisus:
Okay cool I'll take a look at that.

Magisus:
So as far as I can tell, using `set` does not persist across uses of the macro from different projects, even if you cache (which makes sense, the cache file is local to the project). Environment vars are also set for the current run only. So, Facter will always overwrite the reference to `CMAKE_INSTALL_PREFIX` with its own install location, unless I'm missing something.

MikaelSmith:
Oh, I think I misread what that was doing. There are probably some other ideas around using https://github.com/puppetlabs/leatherman/blob/master/LeathermanConfig.cmake.in.

PR #240, opened 2017-05-01
(LTH-124) Update for Ruby 2.4 unification of Fixnum and Bignum

Magisus:
Ruby 2.4 unified the Fixnum and Bignum types into a single Integer type.
This adjusts Leatherman's Ruby API to take that into account.

NOTE: this will require some updates in Facter, since it is an API breaking change.


McdonaldSeanp:
👍

MikaelSmith:
rb_cInteger appears to have been around since at least Ruby 2.1. AppVeyor tests with Ruby 2.1. So 👍

PR #246, opened 2017-08-12
(PE-21053) Enable nxcompat and dynamicbase flags in Windows builds

jsane:
Being done to meet customer's security audit requirement.


jsane:
Re-submitting with correct ticket number. https://github.com/puppetlabs/leatherman/pull/247.

PR #249, opened 2017-08-16
(LTH-139) Allow setting permissions for exec redirection

MikaelSmith:
Apply file permissions to stdout and stderr files created for the output
of an executed process.

### DIFF
--- a/curl/src/client.cc
+++ b/curl/src/client.cc
@@ -8,7 +8,6 @@
 #include <boost/algorithm/string.hpp>
 #include <boost/nowide/cstdio.hpp>
 #include <sstream>
-#include <cstdio>
 
 // Mark string for translation (alias for leatherman::locale::format)
 using leatherman::locale::_;
@@ -170,7 +169,7 @@ namespace leatherman { namespace curl {
         return res;
     }
 
-    void client::download_file(request const& req, std::string const& file_path)
+    void client::download_file(request const& req, std::string const& file_path, boost::optional<fs::perms> perms)
     {
         response res;
         context ctx(req, res);
@@ -187,6 +186,14 @@ namespace leatherman { namespace curl {
             fp = boost::nowide::fopen(temp_path.string().c_str(), "wb");
             if (!fp) {
                 throw http_file_download_exception(req, file_path, _("Failed to open temporary file for writing"));
+            }
+
+            if (perms) {
+                boost::system::error_code ec;
+                fs::permissions(temp_path.string(), *perms, ec);
+                if (ec) {
+                    throw http_request_exception(req, _("Failed to modify permissions of temporary file"));
+                }
             }
 
             curl_easy_setopt_maybe(ctx, CURLOPT_NOPROGRESS, 1);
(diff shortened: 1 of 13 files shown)


MikaelSmith:
Does this approach seem reasonable for restricting file permissions on creation? Alternatively we could just enforce explicit permissions on the file.

MikaelSmith:
Puppet seems to just set file permission. That seems straightforward enough; I don't have an explicit reason to mess with umask.

MikaelSmith:
Going to simplify this and just set permissions.

MikaelSmith:
This should be rebased and merged after https://github.com/puppetlabs/leatherman/pull/250.

curl/src/client.cc
```diff
@@ -207,34 +214,26 @@ namespace leatherman { namespace curl {
...
 
-            LOG_DEBUG("download completed, now writing result to file");
+            LOG_DEBUG("download completed, now writing result to file {1}", file_path);
             fs::rename(temp_path, file_path);
         } catch (http_file_download_exception& e) {
             // so that the code in the lower catch block is not hit
             throw e;
         } catch (http_request_exception& e) {
-            cleanup_temp_file(temp_path, fp, req, file_path, e.what());
+            fclose(fp);
```
MikaelSmith: @ekinanp I did a bit of cleanup around error handling here.

ekinanp: I see, so http_request_exceptions imply that you need to remove the temp file, while raw file_download_exception means that the temp file is gone or not relevant. Yeah, looks good to me @MikaelSmith

execution/src/execution.cc
```diff
@@ -261,15 +277,30 @@ namespace leatherman { namespace execution {
...
+            throw execution_exception(_("failed to open output file {1}", out_file));
+        }
+
+        boost::system::error_code ec;
+        if (perms) {
+            fs::permissions(out_file, *perms, ec);
+            if (ec) {
+                throw execution_exception(_("failed to modify permissions on output file {1} to {2,num,oct}: {3}", out_file, *perms, ec.message()));
+            }
         }
```
mruzicka: I wish there was a way in C++ to open a file and set its permissions in a single step (e.g. like in C with the [`open`](http://man7.org/linux/man-pages/man2/open.2.html) function). As if done in two steps like here, there is still a short window when someone can open the file before its permissions are set more restrictively.

execution/src/execution.cc
```diff
@@ -261,15 +277,30 @@ namespace leatherman { namespace execution {
...
             setup_execute(stderr_callback, actual_options);
         } else {
             err_stream.open(err_file.c_str(), std::ios::binary);
             if (!err_stream.is_open()) {
-                throw execution_exception(_("failed to open the error file."));
+                throw execution_exception(_("failed to open error file {1}", err_file));
+            }
+
+            if (perms) {
+                fs::permissions(err_file, *perms, ec);
```
mruzicka: I wish there was a way in C++ to open file and set its permissions in a single step (like in C's [`open`](http://man7.org/linux/man-pages/man2/open.2.html) function). As if done in two steps like here, there is still a short window when someone can open the file before its permissions are set more restrictively.

execution/src/execution.cc
MikaelSmith: We could do that. I'm going to release anyway to unblock other work, and we can fix that in a .z if we want.

MikaelSmith: @mruzicka going to leave that up to you if you want to add it. The behavior of creating the file then modifying permissions matches Puppet's, but it would be an improvement to set on creation. And we could use the `open` function directly for Windows.

PR #251, opened 2017-08-17
(LTH-140) Download file from a given URL

ekinanp:
Leatherman.curl has been extended to now download a file from a
provided URL and write the downloaded contents to a user-specified
file path. Some refactoring was also done in Leatherman.curl to
take out common error-checking patterns.

NOTE: I am going to start writing tests for this PR. This is just to show the main logic and the refactoring.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: cpp
+dist: precise
 compiler:
   - gcc
 sudo: false
--- a/curl/inc/leatherman/curl/client.hpp
+++ b/curl/inc/leatherman/curl/client.hpp
@@ -8,6 +8,7 @@
 #include "request.hpp"
 #include "response.hpp"
 #include <curl/curl.h>
+#include <boost/filesystem.hpp>
 #include "export.h"
 
 
@@ -108,6 +109,58 @@ namespace leatherman { namespace curl {
     };
 
     /**
+     * The exception for HTTP file downloads.
+     */
+    struct LEATHERMAN_CURL_EXPORT http_file_download_exception : http_request_exception
+    {
+        /**
+         * Constructs an http_file_download_exception.
+         * @param request The request that caused the exception
+         * @param file_path The file that was meant to be downloaded
+         * @param message The exception message.
+         */
+        http_file_download_exception(request req, std::string file_path, std::string const &message) : http_file_download_exception(req, file_path, "", message)
+        {
+        }
+
+        /**
+         * Constructs an http_file_download_exception.
+         * @param request The request that caused the exception
+         * @param file_path The file that was meant to be downloaded
+         * @param temp_path The path to the temporary file that wasn't successfully cleaned up.
+         * @param message The exception message.
+         */
+        http_file_download_exception(request req, std::string file_path, std::string temp_path, std::string const &message) :
+            http_request_exception(req, message),
+            _file_path(std::move(file_path)),
+            _temp_path(std::move(temp_path))
+        {
+        }
+
+        /**
+         * Gets the file_path associated with the exception
+         * @return Returns the file_path associated with the exception.
+         */
+        std::string const& file_path() const
+        {
+            return _file_path;
+        }
+
+        /**
+         * Gets the temp_path associated with the exception
+         * @return Returns the temp_path associated with the exception.
+         */
+        std::string const& temp_path() const
+        {
+            return _temp_path;
+        }
+
+     private:
+        std::string _file_path;
+        std::string _temp_path;
+    };
+
+    /**
      * Implements a client for HTTP.
      * Note: this class is not thread-safe.
      */
@@ -151,6 +204,13 @@ namespace leatherman { namespace curl {
          * @return Returns the HTTP response.
          */
         response put(request const& req);
+
+        /**
+         * Downloads the file from the specified url.
+         * @param req The HTTP request to perform 
+         * @param file_path The file that the downloaded contents will be written to. 
+         */
+        void download_file(request const& req, std::string const& file_path);
 
         /**
          * Sets the path to the CA certificate file.
@@ -216,6 +276,8 @@ namespace leatherman { namespace curl {
         LEATHERMAN_CURL_NO_EXPORT void set_ca_info(context& ctx);
         LEATHERMAN_CURL_NO_EXPORT void set_client_protocols(context& ctx);
 
+        LEATHERMAN_CURL_NO_EXPORT void cleanup_temp_file(boost::filesystem::path const& temp_path, FILE* tfp, request const& req, std::string const& file_path, std::string const& reason);
+
         template <typename ParamType>
         LEATHERMAN_CURL_NO_EXPORT void curl_easy_setopt_maybe(
             context &ctx,
@@ -232,6 +294,7 @@ namespace leatherman { namespace curl {
         static int seek_body(void* ptr, curl_off_t offset, int origin);
         static size_t write_header(char* buffer, size_t size, size_t count, void* ptr);
         static size_t write_body(char* buffer, size_t size, size_t count, void* ptr);
+        static size_t write_file(char *buffer, size_t size, size_t count, void* ptr);
         static int debug(CURL* handle, curl_infotype type, char* data, size_t size, void* ptr);
 
         curl_handle _handle;
--- a/curl/inc/leatherman/curl/client.hpp
+++ b/curl/inc/leatherman/curl/client.hpp
@@ -216,6 +216,18 @@ namespace leatherman { namespace curl {
         LEATHERMAN_CURL_NO_EXPORT void set_ca_info(context& ctx);
         LEATHERMAN_CURL_NO_EXPORT void set_client_protocols(context& ctx);
 
+        template <typename ParamType>
+        LEATHERMAN_CURL_NO_EXPORT void curl_easy_setopt_maybe(
+            context &ctx,
+            CURLoption option,
+            ParamType const& param
+        ) {
+            auto result = curl_easy_setopt(_handle, option, param);
+            if (result != CURLE_OK) {
+                throw http_request_exception(ctx.req, curl_easy_strerror(result));
+            }
+        }
+
         static size_t read_body(char* buffer, size_t size, size_t count, void* ptr);
         static int seek_body(void* ptr, curl_off_t offset, int origin);
         static size_t write_header(char* buffer, size_t size, size_t count, void* ptr);
--- a/curl/src/client.cc
+++ b/curl/src/client.cc
@@ -133,14 +133,8 @@ namespace leatherman { namespace curl {
         curl_easy_reset(_handle);
 
         // Set common options
-        auto result = curl_easy_setopt(_handle, CURLOPT_NOPROGRESS, 1);
-        if (result != CURLE_OK) {
-            throw http_request_exception(req, curl_easy_strerror(result));
-        }
-        result = curl_easy_setopt(_handle, CURLOPT_FOLLOWLOCATION, 1);
-        if (result != CURLE_OK) {
-            throw http_request_exception(req, curl_easy_strerror(result));
-        }
+        curl_easy_setopt_maybe(ctx, CURLOPT_NOPROGRESS, 1);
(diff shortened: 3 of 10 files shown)


curl/src/client.cc
```diff
@@ -296,63 +314,54 @@ namespace leatherman { namespace curl {
...
-                throw http_request_exception(ctx.req, curl_easy_strerror(result));
-            }
+        if (_ca_cert == "") {
+            return;
         }
+        
+        curl_easy_setopt_maybe<const char*>(ctx, CURLOPT_CAINFO, _ca_cert.c_str());
+    }
+
+    void client::set_ca_info(std::string const& url, std::string const& file_path){
```
MikaelSmith: What is this doing? It makes no sense to me.

curl/inc/leatherman/curl/client.hpp
```diff
@@ -213,13 +262,41 @@ namespace leatherman { namespace curl {
         LEATHERMAN_CURL_NO_EXPORT void set_timeouts(context& ctx);
         LEATHERMAN_CURL_NO_EXPORT void set_write_callbacks(context& ctx);
         LEATHERMAN_CURL_NO_EXPORT void set_client_info(context &ctx);
+        LEATHERMAN_CURL_NO_EXPORT void set_client_info(std::string const& url, std::string const& file_path);
         LEATHERMAN_CURL_NO_EXPORT void set_ca_info(context& ctx);
+        LEATHERMAN_CURL_NO_EXPORT void set_ca_info(std::string const& url, std::string const& file_path);
         LEATHERMAN_CURL_NO_EXPORT void set_client_protocols(context& ctx);
 
+        template <typename ParamType> 
+        LEATHERMAN_CURL_NO_EXPORT void curl_easy_setopt_maybe(
```
MikaelSmith: Refactoring around this is better done as a separate commit.

curl/src/client.cc
ekinanp: I can one line it, was a wrapper to curl_easy_setopt_maybe. I wanted to preserve the existing pattern in perform where they set_ca_info, then set_client_info, etc.

curl/inc/leatherman/curl/client.hpp
ekinanp: Do you want me to make it a separate commit? I can make the refactoring for the main code base (before the download_file) as one commit, then add the download_file as another commit.

curl/src/client.cc
MikaelSmith: Oh, it's a helper for the exception that will be thrown. That's somewhat confusing. I'd be ok with set_ca_info just throwing one type of exception.

curl/inc/leatherman/curl/client.hpp
MikaelSmith: Yes please.

curl/inc/leatherman/curl/client.hpp
```diff
@@ -108,6 +108,48 @@ namespace leatherman { namespace curl {
...
+     */
+    struct LEATHERMAN_CURL_EXPORT http_file_download_exception : http_exception
+    {
+        /**
+         * Constructs an http_file_download_exception.
+         * @param url The url that caused the exception.
+         * @param file_path The file that was meant to be downloaded
+         * @param message The exception message.
+         */
+        http_file_download_exception(std::string const &url, std::string const &file_path, std::string const &message) :
```
MikaelSmith: The standard pattern for construction is to pass by value, then move into the object. You're mixing them by passing by reference, which is causing an extra copy.

curl/src/client.cc
MikaelSmith: I think this is mostly a case of non-obvious behavior. I'd make the name clearer about what it's doing.

curl/src/client.cc
```diff
@@ -133,14 +138,8 @@ namespace leatherman { namespace curl {
...
         // Set common options
-        auto result = curl_easy_setopt(_handle, CURLOPT_NOPROGRESS, 1);
-        if (result != CURLE_OK) {
-            throw http_request_exception(req, curl_easy_strerror(result));
-        }
-        result = curl_easy_setopt(_handle, CURLOPT_FOLLOWLOCATION, 1);
-        if (result != CURLE_OK) {
-            throw http_request_exception(req, curl_easy_strerror(result));
-        }
+        curl_easy_setopt_maybe<long>(ctx, CURLOPT_NOPROGRESS, 1);
```
MikaelSmith: The template parameter should be inferred from the arguments. No need to specify the `<long>`. Should be able to avoid defining all the callback signatures as well.

locales/leatherman.pot
```diff
@@ -18,32 +18,51 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 
 #: curl/src/client.cc
+#: curl/src/client_with_download.cc
```
MikaelSmith: Picked up some extra entries. Looks better though.

ekinanp: Yeah, those should go away once I get the download stuff up there.

MikaelSmith:
Couple issues with cpplint, and something funny's going on with the Windows build. I dig more tomorrow.

ekinanp:
Ok put up an updated PR with tests. I'm not sure if I should test for all the failure cases of curl_easy_setopt like the existing curl tests do, but I can if that is necessary. I have also tested the download code by making a request to the orchestrator on my master vm and it worked -- the SSL was set up correctly, and once I added the X-Authentication token, the response was downloaded as a file. I'm thinking it might not be a bad idea to add the token as a field to the client class instead of as a parameter to the download method, since it seems like something that'd belong w/ it.

MikaelSmith:
Ah, two Windows issues: we need to add the dependency on Boost.Filesystem to curl/CMakeLists.txt. Append `filesystem system` to the 1st line.

Also tests fail on Windows. Looks like you might be re-using a temporary file.

MikaelSmith:
https://github.com/puppetlabs/orchestrator-client/blob/hoyt/lib/src/client.cc#L122 is how orchestrator-client adds the X-Auth token. I'm surprised that's needed for the static file download though.

curl/src/client.cc
```diff
@@ -173,6 +171,65 @@ namespace leatherman { namespace curl {
...
+            curl_easy_setopt_maybe(url, file_path, CURLOPT_WRITEDATA, fp);
+
+            // TODO: How long should connection time be? Is 5 minutes too much?
+            curl_easy_setopt_maybe(url, file_path, CURLOPT_CONNECTTIMEOUT, 300);
+            set_ca_info(url, file_path);
+            set_client_info(url, file_path);
+
+            // Authentication header
+            curl_list x_auth_header;
+            x_auth_header.append("X-Authentication: " + token);
```
MikaelSmith: This shouldn't be necessary. It can also be handled externally with add_headers.

curl/src/client.cc
```diff
@@ -173,6 +171,65 @@ namespace leatherman { namespace curl {
...
+            auto fp = boost::nowide::fopen(temp_path.string().c_str(), "wb");
+            if (!fp) {
+                throw http_file_download_exception(url, file_path, _("Failed to open temporary file for writing"));
+            }
+
+            curl_easy_setopt_maybe(url, file_path, CURLOPT_WRITEFUNCTION, write_file);
+            curl_easy_setopt_maybe(url, file_path, CURLOPT_WRITEDATA, fp);
+
+            // TODO: How long should connection time be? Is 5 minutes too much?
+            curl_easy_setopt_maybe(url, file_path, CURLOPT_CONNECTTIMEOUT, 300);
```
MikaelSmith: This is handled by the caller via set_timeouts.

curl/src/client.cc
```diff
@@ -173,6 +171,65 @@ namespace leatherman { namespace curl {
         return res;
     }
 
+    void client::download_file(std::string const& url, std::string const& file_path, std::string const& token)
```
MikaelSmith: I'd have the signature take a request object, and a file_path.

ekinanp:
If I have a printf statement in there, that's to help me see where Travis fails. For some reason, the Windows part in appveyor is passing, but Travis release is getting a segfault at the fopen failure test, so I've been trying to figure out why. Anyways TL;DR, the printf statement will be removed.

curl/src/client.cc
```diff
@@ -6,12 +6,16 @@
 #include <leatherman/locale/locale.hpp>
 #include <boost/utility/string_ref.hpp>
 #include <boost/algorithm/string.hpp>
+#include <boost/filesystem.hpp>
+#include <boost/nowide/cstdio.hpp>
 #include <sstream>
+#include <stdio.h>
```
MikaelSmith: Is this for `fclose`? I'd prefer including via `cstdio` for C++.

curl/src/client.cc
```diff
@@ -173,6 +171,79 @@ namespace leatherman { namespace curl {
...
+            if (result == CURLE_OK) {
+                  printf("Error was before download.\n");
+              throw http_file_download_exception(req, file_path, e.what());
+            }
+                 printf("Error was in temp file removal\n"); 
+
+            throw http_file_download_exception(
+                req,
+                file_path,
+                errbuf + _(" and failed to remove temporary file"));
```
MikaelSmith: This should use string substitution if we want it to translate usefully. As in `_("{1} and failed to remove temporary file", errbuf)`.

curl/src/client.cc
```diff
@@ -173,6 +171,79 @@ namespace leatherman { namespace curl {
...
+
+            LOG_DEBUG("download completed, now writing result to file");
+            fs::rename(temp_path, file_path);
+        } catch (http_file_download_exception& e) {
+                 printf("Threw a file_download exception, re-throwing it. Message is: %s\n", e.what());
+            // so that the code in the lower catch block is not hit
+            throw e;
+        } catch (http_request_exception& e) {
+                 printf("Threw a request exception, converting to download. Message is: %s\n", e.what());
+            fclose(fp);
```
MikaelSmith: This should also ensure the file is removed.

ekinanp: You're right, making the fix now.

.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: cpp
+dist: precise
```
MikaelSmith: Seems like this wasn't needed to fix the test.

curl/inc/leatherman/curl/client.hpp
```diff
@@ -153,6 +206,13 @@ namespace leatherman { namespace curl {
         response put(request const& req);
 
         /**
+         * Downloads the file from the specified url.
+         * @param req The HTTP request to perform 
+         * @param file_path The file that the downloaded contents will be written to. 
+         */
+        void download_file(request const& req, std::string const& file_path);
```
MikaelSmith: Since we're writing a file, we should probably allow specifying the permissions as well. I can handle that in https://github.com/puppetlabs/leatherman/pull/249 as well if you'd prefer.

curl/src/client.cc
```diff
@@ -173,6 +170,85 @@ namespace leatherman { namespace curl {
         return res;
     }
 
+    void client::download_file(request const& req, std::string const& file_path)
+    {
+                   printf("Entered download_file\n");
```
MikaelSmith: Lots of printf's to remove.

.travis.yml
MikaelSmith: Turns out I was wrong, we'd just been getting Precise builds until now.

curl/inc/leatherman/curl/client.hpp
```diff
@@ -216,10 +276,25 @@ namespace leatherman { namespace curl {
         LEATHERMAN_CURL_NO_EXPORT void set_ca_info(context& ctx);
         LEATHERMAN_CURL_NO_EXPORT void set_client_protocols(context& ctx);
 
+        LEATHERMAN_CURL_NO_EXPORT void cleanup_temp_file(boost::filesystem::path const& temp_path, FILE* tfp, request const& req, std::string const& file_path, std::string const& reason);
```
MikaelSmith: I'm going to clean it up separately, but this could be done as a forward-declaration rather than including the header file.

PR #254, opened 2017-08-24
(LTH-142) Separate file operation errors from server-side errors

ekinanp:
Previously, file operation errors (e.g. setting permissions,
writing to disk) were encapsulated the same way as server-side
errors in a single http_file_download_exception. This commit
creates a new exception for file operation errors,
http_file_operation_exception, to distinguish these cases.

### DIFF
--- a/curl/inc/leatherman/curl/client.hpp
+++ b/curl/inc/leatherman/curl/client.hpp
@@ -5,11 +5,13 @@
 #pragma once
 
 #include <leatherman/util/scoped_resource.hpp>
+#include <leatherman/locale/locale.hpp>
 #include "request.hpp"
 #include "response.hpp"
 #include <curl/curl.h>
 #include <boost/optional.hpp>
 #include <boost/filesystem.hpp>
+#include <boost/nowide/cstdio.hpp>
 #include "export.h"
 
 
@@ -66,6 +68,40 @@ namespace leatherman { namespace curl {
     };
 
     /**
+     * Resource for a temporary file used during download
+     */
+    struct LEATHERMAN_CURL_NO_EXPORT download_temp_file
+    {
+        /**
+         * Constructs a temporary file that will be used to store the downloaded file's
+         * contents.
+         * @param req The HTTP request.
+         * @param file_path The file path that this temporary file's contents will be written to.
+         * @param perms The (optional) permissions of the downloaded file.
+         */
+        download_temp_file(request const& req, std::string const& file_path, boost::optional<boost::filesystem::perms> perms);
+
+        ~download_temp_file();
+
+        /*
+         * Returns the underlying file pointer.
+         */
+        FILE* get_fp();
+
+        /*
+         * Writes the temporary file's contents to its file path.
+         */
+        void write();
+
+     private:
+        void cleanup();
+        FILE* _fp;
+        request _req;
+        std::string _file_path;
+        boost::filesystem::path _temp_path;
+    };
+
+    /**
      * The exception for HTTP.
      */
     struct LEATHERMAN_CURL_EXPORT http_exception : std::runtime_error
@@ -110,31 +146,50 @@ namespace leatherman { namespace curl {
     };
 
     /**
-     * The exception for HTTP file downloads.
+     * The exception for curl_easy_setopt errors.
      */
-    struct LEATHERMAN_CURL_EXPORT http_file_download_exception : http_request_exception
+    struct LEATHERMAN_CURL_EXPORT http_setup_exception : http_request_exception
     {
         /**
-         * Constructs an http_file_download_exception.
-         * @param request The request that caused the exception
-         * @param file_path The file that was meant to be downloaded
+         * Constructs an http_setup_exception.
+         * @param req The HTTP request that caused the exception.
          * @param message The exception message.
+         * @param curl_opt The CURL option that failed.
          */
-        http_file_download_exception(request req, std::string file_path, std::string const &message) : http_file_download_exception(req, file_path, "", message)
+        http_setup_exception(request req, CURLoption curl_opt, std::string const &message) :
+            http_request_exception(req, message),
+            _curl_opt(std::move(curl_opt))
         {
         }
 
         /**
-         * Constructs an http_file_download_exception.
+         * Gets the CURL option associated with the exception
+         * @return Returns the CURL option associated with the exception.
+         */
+        CURLoption const& curl_opt() const
+        {
+            return _curl_opt;
+        }
+
+
+     private:
+        CURLoption _curl_opt;
+    };
+
+    /**
+     * The exception for HTTP file download server-side errors.
+     */
+    struct LEATHERMAN_CURL_EXPORT http_download_exception : http_request_exception
+    {
+        /**
+         * Constructs an http_download_exception.
          * @param request The request that caused the exception
          * @param file_path The file that was meant to be downloaded
-         * @param temp_path The path to the temporary file that wasn't successfully cleaned up.
          * @param message The exception message.
          */
-        http_file_download_exception(request req, std::string file_path, std::string temp_path, std::string const &message) :
-            http_request_exception(req, message),
-            _file_path(std::move(file_path)),
-            _temp_path(std::move(temp_path))
+        http_download_exception(request req, std::string file_path, std::string const &message) :
+          http_request_exception(req, message),
+          _file_path(std::move(file_path))
         {
         }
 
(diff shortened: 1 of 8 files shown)


curl/tests/client_test.cc
```diff
@@ -487,15 +491,30 @@ namespace leatherman { namespace curl {
...
-         SECTION("when curl_easy_perform fails and removal of the temporary file fails, then both the errbuf and the failed temporary file removal message are contained in the thrown http_file_download_exception") {
+         SECTION("when curl_easy_perform fails and removal of the temporary file fails, then both the errbuf and the failed temporary file removal message are contained in the thrown http_file_operation_exception") {
              std::string file_path = (temp_dir_path / "file").string();
              curl::request req("");
              test_impl->test_failure_mode = curl_impl::error_mode::easy_perform_error; 
              test_impl->trigger_external_failure = setup_temp_removal_failure;
              REQUIRE_THROWS_WITH(test_client.download_file(req, file_path), Catch::StartsWith("easy perform failed and failed to remove temporary file file_util_fixture_")); 
          }
+
+         SECTION("when curl_easy_perform fails due to a CURLE_WRITE_ERROR, an http_file_operation_exception is thrown") {
```
ekinanp: I'm wondering if for tests such as these, we should either:

(1) Write file_download in a try-catch block w/ an http_file_operation_exception in the catch, and have a statement in there that requires the message to match what we want or

(2) Create a custom matcher that ensures the thrown exception is the right type and it contains the message we want.

From what I've seen, Catch doesn't provide a matcher that checks for both, it's either you check the type of exception thrown OR the message but not both. Unless I'm mistaken.

curl/src/client.cc
```diff
@@ -181,17 +181,19 @@ namespace leatherman { namespace curl {
         char errbuf[CURL_ERROR_SIZE] = { '\0' };
         FILE* fp = nullptr;
         fs::path temp_path = fs::path("");
+        bool file_operation_error = false;
         try {
             temp_path = fs::path(file_path).parent_path() / fs::unique_path("temp_file_%%%%-%%%%-%%%%-%%%%");
             fp = boost::nowide::fopen(temp_path.string().c_str(), "wb");
```
MikaelSmith: The logic around closing and cleaning up resources is getting messy to reason about. I'd suggest looking at scoped_resource; maybe use it to close the file and optionally remove it if it exists.

I want to be able to throw the exception I care about, and let RAII handle cleanup.

curl/tests/client_test.cc
```diff
@@ -469,15 +473,15 @@ namespace leatherman { namespace curl {
              test_impl->test_failure_mode = curl_impl::error_mode::set_url_error;
              REQUIRE_THROWS_AS(test_client.download_file(req, file_path), curl::http_file_download_exception);
```
MikaelSmith: This should assert the http_file_operation_exception.

curl/tests/client_test.cc
MikaelSmith: In the past, I think I've done the try-catch and check for message. You have to have a REQUIRE(false) at the end of the try to ensure you don't reach that point.

curl/tests/client_test.cc
ekinanp: That one is when only curl_easy_setopt fails.

MikaelSmith: Oh, I missed that this and line 459/463 were separate sections. Nevermind.

MikaelSmith: Actually, curl_easy_setopt failing is probably a 3rd typo of exception...

ekinanp: Hmm, so an http_curl_setup_exception?

MikaelSmith: Sure. I'd rather this were a flatter taxonomy, and since no one's likely using this I'd be ok breaking it a bit.

I'd see something like
http_exception
- http_request_exception
- http_setup_exception
- http_file_exception

ekinanp: Hmm, what about:

http_exception:
  -http_request_exception
     -http_setup_exception
     -http_download_exception
        -http_file_exception

My reasoning is that the setup failure results from the passed-in parameter at the end of easy_setopt, which comes from the request. So the request is relevant to that. http_file_exception is part of the download and it may result after a download fails (e.g. when we try to remove the temporary file), so it's important that it preserves the context of the original download exception.

MikaelSmith: So if we want to isolate just server connection failures, you'd have to catch http_setup_exception first and ignore it, then catch http_request_exception. Or am I catching the http_download_exception?

I don't see an important distinction between request and download. For file exception I can care what file it was trying to operate on. But if we fail to connect to a server, I don't really care where the file was going to be stored.

ekinanp: Server connection failures would result in two exceptions being thrown:

- http_request
- http_download

so if you want to isolate them, you can just catch either http_request or http_download, ignoring http_setup completely. If you care about whether something went wrong in the setup, then you'd need to catch setup before the request/download.

The main distinction between request and download is that download has a file. If the download_file call were nested deep in a bunch of method calls, and only the top-level caller has the try-catch block, then it is possible that you can't get the path of the file as the caller when download_file throws an exception. That's why I think it would help for the exception to contain that information.

MikaelSmith: Ok, I'll accept that.

curl/src/client.cc
```diff
@@ -93,6 +93,83 @@ namespace leatherman { namespace curl {
...
+             LOG_DEBUG("download completed, now writing result to file {1}", _file_path);
+             fs::rename(_temp_path, _file_path, ec);
+             if (!ec) {
+                 return;
+             }
+             _external_failure_reason = _("Failed to move over the temporary file's downloaded contents");
+             _throw_external_exception = make_file_exception_thrower(_req, _file_path);
+         }
+         // At this point, we do have an external exception to throw.
+         fs::remove(_temp_path, ec);
```
MikaelSmith: I'd probably be ok swallowing this exception. It shouldn't generally happen - we should be able to delete any file we create (testing verifies that we don't do anything to prevent that). If it's missing, I don't care about it.

ekinanp: That makes sense, it is an edge case. I was thinking you could have a situation where the permissions of the directory containing the temp file could somehow change so that the removal might fail (or something else), + it is handled here and is not too much code. But if it is better to swallow the exception instead, should I just ignore the error code then?

And ok, I can add the logging in.

MikaelSmith: It seems like quite a bit of extra exception handling complexity, with all the set_throw_external_exception handling.

ekinanp: If I do remove the error checking of fs::remove for the temp file, I think the set_throw.. will still need to be there because the original, failure causing external exception does need to be thrown once the temp file is cleaned up, but the external_failure_reason part can be removed.

MikaelSmith: You handle that via the destructor. My point was to make sure cleanup doesn't throw an exception - because the exceptions there are unlikely and not really actionable - so you can rely on the destructor.

ekinanp:
@MikaelSmith Alright I made the changes. Didn't make temp_file a scoped resource b/c it's managing two resources, the temporary file path and the file pointer, while scoped_resource seems to be designed for managing one resource. It seemed simpler than dealing w/ the _deleter component of scoped_resource, but I can make it derived from scoped_resource if that is what is preferred.

curl/src/client.cc
```diff
@@ -93,6 +91,54 @@ namespace leatherman { namespace curl {
...
+        fs::rename(_temp_path, _file_path, ec);
+        if (ec) {
+          throw http_file_exception(_req, _file_path, _("Failed to move over the temporary file's downloaded contents"));
+        }
+    }
+
+    void download_temp_file::cleanup() {
+        fclose(_fp);
+        boost::system::error_code ec;
+        fs::remove(_temp_path, ec);
```
MikaelSmith: We could log a debug message here for later debugging if there's an error.

MikaelSmith:
Of course, there's a cppcheck warning and a test failure on Windows.

ekinanp:
@MikaelSmith Okay, I added some logging, fixed the CI issues and reverted my extra commit.

MikaelSmith:
Can you add an extra commit with release prep? Bump to version 1.1.2 in CMakeLists.txt, add a CHANGELOG entry, build to update any generated files. Use the `(packaging)` prefix for it.

ekinanp:
@MikaelSmith Just did it.

curl/inc/leatherman/curl/client.hpp
```diff
@@ -110,31 +146,50 @@ namespace leatherman { namespace curl {
...
+
+
+     private:
+        CURLoption _curl_opt;
+    };
+
+    /**
+     * The exception for HTTP file download server-side errors.
+     */
+    struct LEATHERMAN_CURL_EXPORT http_download_exception : http_request_exception
```
MikaelSmith: @ekinanp We could've avoided making this a breaking change by keeping the name http_file_download_exception. I'm tempted to do that.

PR #256, opened 2017-08-25
(maint) Refactor http_file_operation_exception

ekinanp:
It now inherits from http_request_exception instead of
http_file_download_exception. This makes it explicit in the
exception class which type of error occurred in curl.


MikaelSmith:
We'll need to tag another release. Can you add that prep?

ekinanp:
Done. @MikaelSmith

PR #258, opened 2017-08-30
(LTH-143) Add a uri parsing utility

MikaelSmith:
Supports parsing a URI into protocol, host, port, path, and query.

### DIFF
--- a/util/src/uri.cc
+++ b/util/src/uri.cc
@@ -1,0 +1,75 @@
+#include <leatherman/util/uri.hpp>
+#include <algorithm>
+#include <sstream>
+
+namespace leatherman { namespace util {
+    uri::uri(std::string const& _uri)
+    {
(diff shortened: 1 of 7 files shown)


util/src/uri.cc
```diff
@@ -0,0 +1,75 @@
...
+
+        // get query start
+        auto query_start = std::find(_uri.begin(), uri_end, '?');
+
+        // protocol
+        auto protocol_start = _uri.begin();
+        auto protocol_end = std::find(protocol_start, uri_end, ':');
+
+        if (protocol_end != uri_end) {
+            std::string prot{protocol_end, uri_end};
```
nicklewis: The variable name `prot` seems to imply this is capturing the protocol string, when in fact it's capturing everything *after* the protocol string.

MikaelSmith: Updated.

PR #259, opened 2017-09-03
(PCP-785) Return response in download_file if HTTP status > 400

ekinanp:
Previously, all information about the HTTP response was contained
inside Leatherman.curl's download_file function. Now, the HTTP
response will only be returned for 400+ HTTP statuses.

### DIFF
--- a/curl/inc/leatherman/curl/client.hpp
+++ b/curl/inc/leatherman/curl/client.hpp
@@ -93,7 +93,15 @@ namespace leatherman { namespace curl {
          */
         void write();
 
+        /*
+         * Writes the temporary file's contents to the body of the
+         * given response.
+         * @param response The HTTP response to write the contents to
+         */
+        void write(response& res);
+
      private:
+        void close_fp();
         void cleanup();
         FILE* _fp;
         request _req;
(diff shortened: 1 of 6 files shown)


curl/inc/leatherman/curl/client.hpp
```diff
@@ -310,10 +318,11 @@ namespace leatherman { namespace curl {
          * @param file_path The file that the downloaded contents will be written to.
          * @param perms The file permissions to apply when writing to file_path.
          *              On Windows this only toggles read-only.
+         * @return Returns the HTTP response if the response status is > 400.
          */
-        void download_file(request const& req,
-                           std::string const& file_path,
-                           boost::optional<boost::filesystem::perms> perms = {});
+        boost::optional<response> download_file(request const& req,
```
MikaelSmith: This doesn't preserve the existing binary API, which we've claimed in the past we'll try to preserve. I'd opt for adding a new function call variant that takes a `boost::optional<response>` reference.

ekinanp:
@MikaelSmith Just addressed your comments.

curl/src/client.cc
```diff
@@ -255,12 +282,22 @@ namespace leatherman { namespace curl {
...
-          throw http_file_operation_exception(req, file_path, make_file_err_msg(_("failed to write to the temporary file during download")));
+            throw http_file_operation_exception(req, file_path, make_file_err_msg(_("failed to write to the temporary file during download")));
         } else if (result != CURLE_OK) {
-          throw http_file_download_exception(req, file_path, _("File download server side error: {1}", errbuf));
+            throw http_file_download_exception(req, file_path, _("File download server side error: {1}", errbuf));
+        }
+
+        // Check the status code. If 400+, fill in the response
+        LOG_DEBUG("request completed (status {1}).", _res.status_code());
+        if (_res.status_code() > 400 && res) {
```
MikaelSmith: How does a caller know when there's something to read in response? I was imaging `res` would only be enabled if status code was `>= 400` (also, needs to be `>= 400`, `not > 400`). It looks like we'll always initialize it if provided, so we'll query the status code?

curl/inc/leatherman/curl/client.hpp
```diff
@@ -316,6 +324,20 @@ namespace leatherman { namespace curl {
                            boost::optional<boost::filesystem::perms> perms = {});
 
         /**
+         * Downloads the file from the specified url.
+         * Throws http_file_download_exception if anything goes wrong.
+         * @param req The HTTP request to perform.
+         * @param file_path The file that the downloaded contents will be written to.
+         * @param response The HTTP response. The body will only be included if the response status is > 400.
```
MikaelSmith: Ok, I guess this is sufficient documentation.

MikaelSmith: Should say `>= 400`.

PR #265, opened 2017-09-19
(LTH-130) Use rapidjson 1.1.0

MikaelSmith:
rapidjson is no longer vendored, it must be installed on the system
independently.


MikaelSmith:
Blocking work:
- Travis and AppVeyor need to be updated to install rapidjson (that'll have to be done on downstream projects that use it as well); curling from https://github.com/Tencent/rapidjson/archive/v1.1.0.zip should work in most cases.
- https://github.com/puppetlabs/puppet-agent needs to be updated to provide rapidjson as a component
- Facter and cpp-project-template should have updates to docs to include installing rapidjson as part of dev setup

json_container/CMakeLists.txt
```diff
@@ -1,11 +1,11 @@
 find_package(Boost 1.54 REQUIRED COMPONENTS regex)
+find_package(rapidjson 1.1.0 REQUIRED)
```
smortex: The file is named `RapidJSON` upstream:
https://github.com/Tencent/rapidjson/blob/master/RapidJSON.pc.in

This seems to be causing the failures…

anthraxx:
any change to get this done?

MikaelSmith:
Closed in favor of #289. I'm not actively working on the problem.

PR #268, opened 2017-11-10
(maint) Make execution exceptions more specific and drop error logging

MikaelSmith:
Move towards better exceptions rather than logging errors. This removes
redundancy in logging and allows better debugging in pxp-agent when
sending a PXP error response. May help debug PA-1618.


MikaelSmith:
There's some work to do in windows, but I also eventually want Leatherman to stop using logging and let you register your own logging hooks.

PR #269, opened 2017-11-13
(LTH-149) Handle errors if child has exited before all writing finished

MikaelSmith:
In testing, we periodically see `failed to run: stdin pipe i/o failed: Broken pipe (32)` on AIX. My hypothesis is that this occurs because the child process exits before we've made a final attempt to write 0 bytes to it and close the pipe. Catch EPIPE when writing 0 bytes and consider the pipe closed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [1.2.2]
+
+### Fixed
+- Allow Leatherman.execute calls to opt into allowing tasks to finish without reading stdin (i.e. don't fail when the pipe is closed) via the `allow_stdin_unread` option. This specifically supports pxp-agent's task execution where input may or may not be used. (LTH-149)
+
 ## [1.2.1]
 
 ### Fixed
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.2.2)
-project(leatherman VERSION 1.2.1)
+project(leatherman VERSION 1.2.2)
 
 if (WIN32)
     link_libraries("-Wl,--nxcompat -Wl,--dynamicbase")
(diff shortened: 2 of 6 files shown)


MikaelSmith:
Does that hypothesis seem reasonable? This test at least seems somewhat sane: if we get EPIPE but we've written everything we expect to, then we don't really care about the error.

branan:
That *seems* reasonable to me. Would it be possible to create an ad-hoc build with this and run a flaky test in a loop for a while to try to confirm if this fixes it?

MikaelSmith:
Sure, I'll build and try running AIX a bunch.

MikaelSmith:
I still saw a failure after recompiling and using this code. So probably not sufficient. I'm working on a new build that incorporates the other PR to improve error messages and will run that for awhile.

MikaelSmith:
Oh, makes total sense why this could happen. The shell script in the test never tries to read stdin, so it could definitely exit before we try to write stdin. How we should handle that is up in the air... if the pipe is closed should we ignore it or propagate the error as a new exception type and let the caller handle it?

MikaelSmith:
Updated to not fail on closed stdin pipe, as it's assumed the executed command will report an error if it didn't get the input it needed.

branan:
I feel like if a program is not expecting input, we shouldn't be sending it. And conversely, if  we send it, we should expect it to be consumed. If we're sending input that we don't expect to be read, that's a bug in our program. If we are expecting it to be read, that's a sign the other program exited prematurely for some reason.

If pxp-agent has *specific* cases where we're sending input that *might* be read, I think it's better to bubble that up as a separate exception so that the pxp-agent code can handle it appropriately for that situation.

MikaelSmith:
Ok, we can go with that. pxp-agent definitely has use cases where input _might_ be read.

MikaelSmith:
Updated to add a new option.

MikaelSmith:
I think what I'm saying is we can't reliably detect that a script is ignoring input it's given, so what benefit do we get from very occasionally failing when that happens?

MikaelSmith:
@branan are you happy with this PR now? I'll add another commit for release prep.

PR #271, opened 2018-01-24
(packaging) Prepare for version 0.12.3 [no-promote]

branan:
0.12.2 was tagged without bumping the version number. This is
confusing to our automation, which assumes version files are
authoritative.

Moving forward, leatherman's version file will be managed by this new
automation, so we should not encounter issues like this again.


McdonaldSeanp:
👍

PR #277, opened 2018-06-14
(LTH-158) Add proxy support to Curl wrapper

donoghuc:
Add ability to configure proxy using CULROPT_PROXY.
https://curl.haxx.se/libcurl/c/CURLOPT_PROXY.html


MikaelSmith:
I'd like a separate LTH ticket for this.

MikaelSmith:
We also need to retarget it against 1.4.x.

PR #284, opened 2018-10-11
(maint) Add a dynamically-built library tarball to github

branan:
The statically-liked boost in Alpine cannot be used to create a DLL as
Facter requires. This adds a version which is linked dynamically to
allow us to use newer builds of leatherman in Facter.


branan:
This cherry-pick had a conflict, so I'd like to see it pass

branan:
I've confirmed Travis has built the right artifact here

PR #286, opened 2018-11-05
(LTH-160) Fix lengthy closure of open file descriptors

nwitkowski:
Currently leatherman closes each FD after stderr when calling do_exec_child.
On FreeBSD this takes a significant amount of time.
This patch changes the FreeBSD implementation to use the closefrom system call
instead of iterating through all possible FDs.

See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=232538 and @smortex

Co-authored-by: David Blacka <david@blacka.com>

### DIFF
--- a/execution/src/posix/execution.cc
+++ b/execution/src/posix/execution.cc
@@ -314,9 +314,13 @@ namespace leatherman { namespace execution {
         }
 
         // Close all open file descriptors above stderr
+#if defined(__FreeBSD__)
+        closefrom(STDERR_FILENO + 1);
+#else
         for (uint64_t i = (STDERR_FILENO + 1); i < max_fd; ++i) {
             close(i);
         }
+#endif
 
         // Execute the given program; this should not return if successful
         execve(program, const_cast<char* const*>(argv), const_cast<char* const*>(envp));


execution/src/posix/execution.cc
```diff
@@ -314,9 +314,13 @@ namespace leatherman { namespace execution {
         }
 
         // Close all open file descriptors above stderr
+#if defined(__FreeBSD__)
```
smortex: Instead of testing the operating system, we can tell CMake to check whether `closefrom(2)` exist or not in libc.  This would allow leatherman to use this system call when available (at least FreeBSD and OpenBSD have it), and rely on the for loop otherwise.

Adding the following to `execution/CMakeLists.txt` defines `HAS_CLOSEFROM` accordingly:

```cmake
check_library_exists(c closefrom /lib CLOSEFROM_IN_LIBC)
if ("${CLOSEFROM_IN_LIBC}" STREQUAL "1")
    add_compile_definitions(HAS_CLOSEFROM)
endif()
```

branan: Yeah, I think I'd prefer to see the detection done in CMake if possible. It looks like Solaris and other BSDs also have the call, so we should use it wherever it's available.

branan:
Closing in favor of #291

PR #288, opened 2018-12-04
(maint) Add option to enable/disable -Werror

ekinanp:
GCC 8.2 outputs far more warnings than previous GCC versions. Thus, to
avoid blocking dev. work on these warnings, this commit adds the
ENABLE_CXX_WERROR option that disables -Werror by default, but still
gives people the option to enable it for e.g. local development.

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -1,4 +1,5 @@
 include(leatherman)
+defoption(ENABLE_CXX_WERROR "Enables the -Werror compiler option" ON)
 defoption(COVERALLS "Generate code coverage using Coveralls.io" OFF)
 defoption(BOOST_STATIC "Use Boost's static libraries" OFF)
 defoption(CURL_STATIC "Use curl's static libraries" OFF)
--- a/cmake/cflags.cmake
+++ b/cmake/cflags.cmake
@@ -2,7 +2,10 @@
 # Each of our project dirs sets CMAKE_CXX_FLAGS based on these. We do
 # not set CMAKE_CXX_FLAGS globally because gtest is not warning-clean.
 if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "\\w*Clang")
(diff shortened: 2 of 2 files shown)


ekinanp:
I've been able to build Fedora 29 for 5.5.x with these changes.

geoffnichols:
fyi @mihaibuzgau

cmake/options.cmake
```diff
@@ -1,4 +1,5 @@
 include(leatherman)
+defoption(ENABLE_CXX_WERROR "Enables the -Werror compiler option" OFF)
```
ekinanp: Don't know if this is the best name for this option.

MikaelSmith: Name seems fine. We're defaulting to off? I'm not keen on ignoring warnings. In the past we've added exceptions where they made sense, or used platform-specific CMAKE_CXX_FLAGS config.

ekinanp: Yeah that's something I wasn't sure about. The reason I defaulted to `OFF` here is to avoid having to set it to `OFF` in all of our components that compile with GCC 8. I'm happy to change it to `ON` if that is preferred. @branan Do you have any thoughts?

branan: @MikaelSmith now that we're building with system toolchains, we're hitting new warnings during pretty much every platform enablement. I think defaulting to off is better for our build environments.

We could possibly change the default based on debug vs release mode, so builds on laptops still have Werror without needing to opt-in?

MikaelSmith: I was thinking more inline with default to on, but we can set this to off in puppet-agent.

MikaelSmith: That way we preserve the checks local development and Travis CI are doing.

ekinanp: I'm going to go ahead and do what @MikaelSmith suggests since we need to get this in for Fedora 29.

PR #289, opened 2019-01-10
(LTH-130) Unvendor RapidJSON

smortex:
This is an attempt of extracting RapidJSON from the leatherman archive.

Work was started in #265 by @MikaelSmith but not finished.  Take his work, rebase on top of `master`, add a fix and see what happens :wink:


smortex:
Yay! After a few hours of pain, it's all green :relieved: 

@MikaelSmith, because of divergences, I had to change the way RapidJSON is made available during the rebase.  For build on UNIX-like systems, I chose to install the `rapidjson-dev` package which contains all required parts:

https://github.com/puppetlabs/leatherman/blob/d83c2a22f534c013a5bf83a2f86d05a6bb564bcd/.travis.yml#L11

For windows, I chose to grab the source code, and use the "regular" (?) build tools to do what I think is equivalent to `./configure && make && make install`:

https://github.com/puppetlabs/leatherman/blob/d83c2a22f534c013a5bf83a2f86d05a6bb564bcd/appveyor.yml#L16-L23

Please tell me if you mind, and I'll try to amend this commit to separate what you wrote from what I wrote.

The other changes are I guess straightforward.

**This is ready for review**, and could replace #265 and #287!

appveyor.yml
```diff
@@ -13,6 +13,15 @@ init:
       choco install -y pl-boost-x64 -Version 1.58.0.2 -source https://www.myget.org/F/puppetlabs
       choco install -y pl-openssl-x64 -Version 1.0.24.1 -source https://www.myget.org/F/puppetlabs
       choco install -y pl-curl-x64 -Version 7.46.0.1 -source https://www.myget.org/F/puppetlabs
+      cd C:\projects
+      curl --location -o v1.1.0.zip https://github.com/Tencent/rapidjson/archive/v1.1.0.zip
+      7z.exe x v1.1.0.zip
+      cd rapidjson-1.1.0
+      cmake .
+      dir
```
smortex: :unamused:

MikaelSmith:
These changes make sense. They have some implications for downstream users. I think we (Puppet) will want to add a pre-compiled rapidjson to our myget feed.

smortex:
> I think we (Puppet) will want to add a pre-compiled rapidjson to our myget feed.

That look like a good idea, however, I don't know how to help to make this happen.

Similarly, adding `rapidjson-dev` directly to the docker image used may make sense.  I found [what I suspect to be the Dockerfile of this image](https://github.com/puppetlabs/cpp-project-template/blob/master/docker/Dockerfile) yesterday, if you think it makes sense, I can open a Pull-Request to do so?

Thanks!

PR #291, opened 2019-02-14
(LTH-160) Use closefrom(2) when available

smortex:
This is a follow-up to #286 by @nwitkowski and @dblacka (I have no permission to push in nwitkowski's repo but have the expected code locally).

I added a commit checking for `closeform(2)` in the CMake configuration and using the new macro instead of checking for the os name.


branan:
Oh, nice! I was gonna do this myself in the next couple days if the original contributor didn't follow up. Thanks so much!

caseywilliams:
Thanks @smortex! :smile:

PR #293, opened 2019-05-29
(maint) Add missing pointer reference to string

McdonaldSeanp:
cenv.cpp has a line that references a char* without the preceeding *, which
will break builds with newer GCC (TBH I don't know how it worked with older
GCC either?).

This commit simply adds the single *


MikaelSmith:
Fix at https://github.com/artyom-beilis/nowide/pull/39/files, but no one seems to be reviewing them.

PR #298, opened 2019-08-02
(LTH-163) Call ruby_sysinit to initialize Windows sockets

GabrielNagy:
Ruby >= 2.5 has changed how the Windows sockets are started (see: ruby/ruby@e33b1690), requiring a call to `rb_w32_sysinit` for starting them.

Not having this call causes segmentation faults when querying custom facts that use Windows Sockets with Ruby versions 2.5 and higher.

### DIFF
--- a/ruby/src/api.cc
+++ b/ruby/src/api.cc
@@ -112,6 +112,7 @@ namespace leatherman { namespace ruby {
         LOAD_SYMBOL(rb_eLoadError),
         LOAD_OPTIONAL_SYMBOL(ruby_setup),
         LOAD_SYMBOL(ruby_init),
+        LOAD_SYMBOL(ruby_sysinit),
         LOAD_SYMBOL(ruby_options),
         LOAD_SYMBOL(ruby_cleanup),
         _library(move(library))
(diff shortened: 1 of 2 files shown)


ruby/src/api.cc
```diff
@@ -157,13 +158,13 @@ namespace leatherman { namespace ruby {
...
-                "ruby",
-                "-e",
-                ""
+            // Calling ruby_sysinit to initialize windows sockets due to the change in
+            // https://github.com/ruby/ruby/commit/e33b1690d06f867e45750bd8e3e8b06d78b5bc26
+            int sysinit_opts_size = 1;
+            char const* sysinit_opts[] = {
+                "ruby"
             };
+            ruby_sysinit(&sysinit_opts_size, reinterpret_cast<char***>(&sysinit_opts));
```
GeorgeMrejea: If this modification works only for ruby version >= 2.5.x, I think it should do a version check or at least documented.

GabrielNagy: This should work regardless of ruby version. Also, this leatherman branch uses puppet-runtime 6.0.x which has ruby-2.5.3: https://github.com/puppetlabs/puppet-runtime/blob/master/configs/projects/agent-runtime-6.0.x.rb#L3

puppet-agent <= 5.5.x will not have this leatherman fix, since this issue is not present there, so we're good.

PR #301, opened 2019-08-22
(FACT-1919) Add environment::get_int to get environment variables as integers

GabrielNagy:
Add a new helper function to simplify getting integer environment variables from the system.

The function parameters are the environment variable name, and the default value to return. If the environment variable to get does not exist, or cannot be cleanly coverted to an integer, the default value is returned instead.

Example usage:
```cpp
static const unsigned int EC2_SESSION_TIMEOUT = environment::get_int("EC2_SESSION_TIMEOUT", 5000);
```

### DIFF
--- a/util/inc/leatherman/util/environment.hpp
+++ b/util/inc/leatherman/util/environment.hpp
@@ -16,6 +16,14 @@ namespace leatherman { namespace util {
     struct environment
     {
         /**
+         * Gets an environment variable as integer.
+         * @param name The name of the environment variable to get.
+         * @param default_value The value to return if the variable is absent or invalid.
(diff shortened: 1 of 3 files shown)


util/inc/leatherman/util/environment.hpp
```diff
@@ -15,6 +15,14 @@ namespace leatherman { namespace util {
      */
     struct environment
     {
+        /**
+         * Gets an environment variable as integer.
+         * @param name The name of the environment variable to get.
+         * @param default_value The value to return if the variable is absent.
```
ciprianbadescu: ... absent or invalid

PR #305, opened 2019-09-13
(LTH-166) Check if waitpid was merely interrupted

aduitsis:
A signal may interrupt a system call such as waitpid, in which case it would return -1 but with errno set to EINTR. The typical solution is to run waitpid inside a while loop and continue looping while waitpid returns -1 with the errno variable is set to EINTR. This patch does exactly that.

This patch fixes this particularly nasty bug: https://tickets.puppetlabs.com/browse/FACT-2019, but one could easily imagine other strange bugs (or rather, Heisenbugs!!!) occurring as a result.

### DIFF
--- a/execution/src/posix/execution.cc
+++ b/execution/src/posix/execution.cc
@@ -503,9 +503,14 @@ namespace leatherman { namespace execution {
                 kill(-child, SIGKILL);
             }
             // Wait for the child to exit
-            if (waitpid(child, &status, 0) == -1) {
-                LOG_DEBUG(format_error(_("waitpid failed")));
-                return;
+            while (waitpid(child, &status, 0) == -1) {
+                if ( errno == EINTR ) {
+                    LOG_DEBUG(format_error(_("waitpid was interrupted by a signal, retrying")));
+                    continue;
+                } else {
+                    LOG_DEBUG(format_error(_("waitpid failed")));
+                    return;
+                }
             }
             if (WIFEXITED(status)) {
                 status = static_cast<char>(WEXITSTATUS(status));


GabrielNagy:
Hi @aduitsis,

Thanks for your contribution! I've changed the project of the ticket you opened to Leatherman, since the fix is being done here.

Two things:
- please retarget your PR on the 1.4.x branch, so this gets in all our [release streams](https://github.com/puppetlabs/leatherman/blob/master/CONTRIBUTING.md#making-changes)
- please amend your commit title to include the ticket name (i.e. `(LTH-166) Check if waitpid was merely interrupted`)

Thank you!

aduitsis:
> Hi @aduitsis,
> 
> Thanks for your contribution! I've changed the project of the ticket you opened to Leatherman, since the fix is being done here.
> 
> Two things:
> 
> * please retarget your PR on the 1.4.x branch, so this gets in all our [release streams](https://github.com/puppetlabs/leatherman/blob/master/CONTRIBUTING.md#making-changes)
> * please amend your commit title to include the ticket name (i.e. `(LTH-166) Check if waitpid was merely interrupted`)
> 
> Thank you!

Hello,

Changed the name, retargeted to branch 1.4.x. Note, I took it from the master and sent it to 1.4.x, is that okay?

Thanks!

allanjude:
I can confirm this issue with leatherman was breaking facter and puppet on FreeBSD with ruby 2.6. Applying this patch resolves the issue.

More detail and debug output in our issue: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=240566

PR #311, opened 2019-10-31
(packaging) Bump to version '1.10.0' [no-promote]

gimmyxd:
1.9.0 was released without running the pipeline, so a manual bump to next version is needed.


GabrielNagy:
Can we bump just the Z? We've bumped Y a lot lately and I don't think leatherman changes are that significant in general.

PR #312, opened 2019-11-05
Fact 2054 facter execution execute expands shell builtins

GeorgeMrejea:


ruby/inc/leatherman/ruby/api.hpp
```diff
@@ -356,6 +356,10 @@ namespace leatherman {  namespace ruby {
          */
         VALUE (* const rb_hash_lookup2)(VALUE, VALUE, VALUE);
         /**
+		 * See MRI documentation.
+		 */
+		VALUE (* const rb_hash_lookup3)(VALUE, bool, VALUE, bool);
```
GeorgeMrejea: compilation problem with rb_hash_lookup3 caused by its signature

ruby/inc/leatherman/ruby/api.hpp
```diff
@@ -356,6 +356,10 @@ namespace leatherman {  namespace ruby {
          */
         VALUE (* const rb_hash_lookup2)(VALUE, VALUE, VALUE);
         /**
+		 * See MRI documentation.
+		 */
+		VALUE (* const rb_hash_lookup3)(VALUE, VALUE, VALUE, VALUE);
```
GeorgeMrejea: compilation problem with rb_hash_lookup3 caused by its signature

GabrielNagy: `rb_hash_lookup3` is not a function MRI implements.

`rb_hash_lookup2` is the same as `rb_hash_lookup` except it returns a custom default value if they key does not exist: https://github.com/ruby/spec/commit/bcb70e2eb309b7a8f17b42cd3f45c1d3c8f87e96

execution/src/posix/execution.cc
```diff
@@ -139,8 +146,37 @@ namespace leatherman { namespace execution {
...
+            char buffer[max_buffer];
+            FILE *stream = popen(cmd.c_str(), "r");
+            if (stream) {
+                while (!feof(stream)) {
+                    if (fgets(buffer, max_buffer, stream) != NULL)
+                        data.append(buffer);
+                }
+                pclose(stream);
+            }
+            builtin = boost::algorithm::contains(data, "builtin");
```
GabrielNagy: I think we can just use `string::find` here.

execution/src/posix/execution.cc
```diff
@@ -139,8 +146,37 @@ namespace leatherman { namespace execution {
...
+        } catch (exception& e) {
+            std::cout << e.what() << std::endl;
+        }
+        return builtin;
+}
```
GabrielNagy: nit: missing indentation

execution/src/posix/execution.cc
```diff
@@ -139,8 +146,37 @@ namespace leatherman { namespace execution {
         return fs.st_mode & S_IXOTH;
     }
 
-    string which(string const& file, vector<string> const& directories)
+    bool isBuiltin(string const& file)
```
GabrielNagy: can we use snake case for the function name? i.e. `is_builtin`
this seems to be the standard here

execution/src/posix/execution.cc
```diff
@@ -26,8 +31,10 @@ using namespace leatherman::util::posix;
 using namespace leatherman::execution;
 using namespace leatherman::logging;
 using namespace leatherman::file_util;
+using namespace boost::algorithm;
 using namespace boost::filesystem;
 
+
```
GabrielNagy: nit: empty newline

execution/src/posix/execution.cc
```diff
@@ -139,8 +146,37 @@ namespace leatherman { namespace execution {
...
+            if (stream) {
+                while (!feof(stream)) {
+                    if (fgets(buffer, max_buffer, stream) != NULL)
+                        data.append(buffer);
+                }
+                pclose(stream);
+            }
+            builtin = boost::algorithm::contains(data, "builtin");
+        } catch (exception& e) {
+            std::cout << e.what() << std::endl;
```
GabrielNagy: can we be more specific about which part of the code can raise an exception? can it even raise an exception?

execution/src/posix/execution.cc
```diff
@@ -15,6 +15,11 @@
 #include <fcntl.h>
 #include <signal.h>
 
+#include <iostream>
+#include <stdio.h>
+#include <stdlib.h>
+
+
```
GabrielNagy: We shouldn't need these I think

execution/src/posix/execution.cc
```diff
@@ -158,7 +194,7 @@ namespace leatherman { namespace execution {
         return {};
     }
 
-    // Represents information about a pipe
+        // Represents information about a pipe
```
GabrielNagy: nit: indentation

execution/tests/posix/execution.cc
```diff
@@ -726,3 +726,23 @@ SCENARIO("executing commands with execution::each_line") {
...
+    GIVEN("set expand to true with a built-in executable") {
+        THEN("the executable absolute path is not found") {
+            boost::format fmt = boost::format("cd %1% && ls") % EXEC_TESTS_DIRECTORY;
+            std::string fmtStr = boost::str(fmt);
+            vector<string> lines;
+            string command = expand_command(fmtStr, lines, true );
+            REQUIRE( command == "");
+        }
+    }
+}
```
GabrielNagy: there's already a SCENARIO that tests this function, we can move these tests there for better readability:
https://github.com/puppetlabs/leatherman/blob/b1d1ea2bb8789ea602594acc8a3c8b8a8008ee67/execution/tests/posix/execution.cc#L83

execution/src/posix/execution.cc
ciprianbadescu: LOG_ERROR should be used instead of cout in case of error cases

execution/src/posix/execution.cc
```diff
@@ -139,8 +146,37 @@ namespace leatherman { namespace execution {
...
 
-    string which(string const& file, vector<string> const& directories)
+    bool isBuiltin(string const& file)
+    {
+        string data;
+        bool builtin;
+        try {
+            string cmd = "type ";
+            cmd.append(file);
+            const int max_buffer = 25 + file.length();
```
ciprianbadescu: use a const instead magic numbers

execution/tests/posix/execution.cc
```diff
@@ -726,3 +726,23 @@ SCENARIO("executing commands with execution::each_line") {
         }
     }
 }
+SCENARIO("Toggle expanding command paths behaviour in execution::expand_command") {
```
ciprianbadescu: I think we should need for cases for all combinations is builtin/command + expand true/false

execution/src/posix/execution.cc
GeorgeMrejea: This is the most concise form I have found

execution/tests/posix/execution.cc
GeorgeMrejea: moved

execution/src/posix/execution.cc
GeorgeMrejea: We do need them

execution/src/posix/execution.cc
GabrielNagy: I'd just do `return data.find("builtin") != std::string::npos;` here but to each his own.

Anyway, there's no need for the `builtin` variable.

execution/src/posix/execution.cc
```diff
@@ -139,8 +144,37 @@ namespace leatherman { namespace execution {
         return fs.st_mode & S_IXOTH;
     }
 
-    string which(string const& file, vector<string> const& directories)
+    bool is_Builtin(string const& file)
```
GabrielNagy:
```suggestion
    bool is_builtin(string const& file)
```

snake case should be all lowercase

execution/src/posix/execution.cc
```diff
@@ -139,8 +144,37 @@ namespace leatherman { namespace execution {
...
+        string data;
+        bool builtin;
+        try {
+            string cmd = "type ";
+            cmd.append(file);
+            const int max_buffer = 25 + file.length();
+            char buffer[max_buffer];
+            FILE *stream = popen(cmd.c_str(), "r");
+            if (stream) {
+                while (!feof(stream)) {
```
GabrielNagy: You should not use `while(!feof(...))` here.
See https://faq.cprogramming.com/cgi-bin/smartfaq.cgi?id=1043284351&answer=1046476070

execution/tests/posix/execution.cc
```diff
@@ -726,3 +744,4 @@ SCENARIO("executing commands with execution::each_line") {
         }
     }
 }
+
```
GabrielNagy: nit: empty newline

execution/inc/leatherman/execution/execution.hpp
```diff
@@ -254,15 +254,15 @@ namespace leatherman { namespace execution {
      * @param directories The directories to search.
      * @return Returns the full path or empty if the file could not be found.
      */
-    std::string which(std::string const& file, std::vector<std::string> const& directories = lth_util::environment::search_paths());
+    std::string which(std::string const& file, std::vector<std::string> const& directories = lth_util::environment::search_paths(), bool expand = true);
```
GabrielNagy: can you also update the documentation string above with the new parameter?

execution/inc/leatherman/execution/execution.hpp
```diff
@@ -254,15 +254,15 @@ namespace leatherman { namespace execution {
...
 
     /**
      * Expands the executable in the command to the full path.
      * @param command The command to expand.
      * @param directories The directories to search.
      * @return Returns the expanded command if the executable was found or empty if it was not found..
      */
-    std::string expand_command(std::string const& command, std::vector<std::string> const& directories = lth_util::environment::search_paths());
+    std::string expand_command(std::string const& command, std::vector<std::string> const& directories = lth_util::environment::search_paths(), bool expand = true);
```
GabrielNagy: can you also update the documentation string above with the new parameter?

execution/src/posix/execution.cc
```diff
@@ -15,6 +15,10 @@
 #include <fcntl.h>
 #include <signal.h>
 
+#include <iostream>
+#include <stdio.h>
+#include <stdlib.h>
```
GabrielNagy: I really believe you're not using `iostream` and `stdlib` here.

execution/src/posix/execution.cc
```diff
@@ -139,8 +142,33 @@ namespace leatherman { namespace execution {
...
+        FILE *stream = popen(cmd.c_str(), "r");
+        if (stream != nullptr) {
+            rewind(stream);
+            if (fgets(buffer, max_buffer, stream) != NULL){
+                data.append(buffer);
+            }
+            pclose(stream);
+        }
+
+        return boost::algorithm::contains(data, "builtin");
```
ciprianbadescu: remove `boost::algorithm::` as namespace `boost::algorithm::` is already imported at Line31

execution/src/posix/execution.cc
ciprianbadescu: It looks to compile without, at lease locally on Linux,

PR #318, opened 2019-12-12
(maint) Throw invalid argument error on Windows

GeorgeMrejea:


execution/src/windows/execution.cc
```diff
@@ -74,6 +74,10 @@ namespace leatherman { namespace execution {
 
     string which(string const& file, vector<string> const& directories, bool expand)
     {
+        //On Windows this function is not supported
+        if (!expand) {
+            throw std::invalid_argument("Unsupported argument on Windows");
```
GabrielNagy: Leatherman has some defined exceptions that wrap the standard library ones.
Looks like the standard in this file is to throw `execution_exception`.

Also, maybe we can log the message using `LOG_ERROR`, like the below cases.

execution/src/execution.cc
```diff
@@ -126,7 +126,11 @@ namespace leatherman { namespace execution {
             }
         }
 
-        file = which(file, directories, expand);
+        try{
+            file = which(file, directories, expand);
+        } catch (const std::invalid_argument &ex) {
+            throw ex;
+        }
```
GabrielNagy: I don't think we need this if we're already throwing the exception below, but I might be wrong.

execution/src/windows/execution.cc
GeorgeMrejea: LOG_ERROR -agree
execution_exception should be used for "normal" execution that encounter problems.
std::invalid_argument shows that it is not "normal" execution (on windows )

execution/src/execution.cc
GeorgeMrejea: Intention is to propagate exception to the caller

GabrielNagy: why though? seems redundant since you're just throwing back the same exception. but ok

execution/src/windows/execution.cc
```diff
@@ -74,6 +74,11 @@ namespace leatherman { namespace execution {
 
     string which(string const& file, vector<string> const& directories, bool expand)
     {
+        // On Windows is not supported to turn off command expansion by setting expand to false
```
GabrielNagy: cpplint fails on this line: Line ends in whitespace.  Consider deleting these extra spaces

PR #319, opened 2020-02-05
cmake: Prefix libraries with lib

ignatenkobrain:
Real libraries must be prefixed with `lib` which is not currently the case for leatherman.


GabrielNagy:
Hi,

We considered this some time ago, however the downstream implications would be quite huge (lots of puppetlabs projects depend on leatherman) and we can't invest in making this change right now. It's also a breaking change and it would have to wait for a major release.

PR #323, opened 2020-06-02
Include header for std::invalid_argument

jwakely:
As documented at https://gcc.gnu.org/gcc-10/porting_to.html#cxx the C++ headers in GCC 10 no longer include `<stdexcept>` in places that don't need it. This causes leatherman to fail to compile because it uses `std::invalid_argument` without including the right header.

### DIFF
--- a/util/src/environment.cc
+++ b/util/src/environment.cc
@@ -1,5 +1,6 @@
 #include <leatherman/util/environment.hpp>
 #include <boost/nowide/cenv.hpp>
+#include <stdexcept>
 
 using namespace std;
 


jwakely:
N.B. I haven't signed your CLA, but this change is trivial.

MikaelSmith:
Thanks, I'll take care of getting this merged in.

MikaelSmith:
jenkins please test this

MikaelSmith:
Actually @puppetlabs/night-s-watch, I might need help. Acceptance tests look misconfigured.

GabrielNagy (approved):

PR #325, opened 2020-06-18
(PA-3254) Failing locales causes 'facter -p' to crash

Dorin-Pleava:
If the current system locales fail, any interpolated facter message will
crash with 'Invalid or unsupported charset:ANSI_X3.4-1968 '.
If locales are not disabled, but they fail to get the locales,
leatherman fails as there is no defense to this.

Now, in case this fails, it will do the interpolation using boost::regex
and boost::format, as if locales were disabled.

Depends on: https://github.com/puppetlabs/puppet-agent/pull/1917


GabrielNagy:
Also I think https://github.com/puppetlabs/cpp-hocon/pull/128 needs to be merged first.

PR #328, opened 2020-11-19
(PA-3492) add event log support

ciprianbadescu:


logging/src/windows/logging.cc
```diff
@@ -49,4 +59,64 @@ namespace leatherman { namespace logging {
...
+        if (h_event_log) {
+            int severity = log_level_to_severity(level);
+            if (severity != STATUS_SEVERITY_SUCCESS) {
+                const int category = 0;
+                const int event_id = 1;
+                const wstring w_message(message.begin(), message.end());
+                LPCWSTR p_w_message = w_message.c_str();
+                if (!ReportEvent(h_event_log, severity, category, event_id, NULL, 1,
+                                 0, &p_w_message, NULL)) {
+                    boost::nowide::cout << "ReportEvent failed with " << GetLastError()
```
McdonaldSeanp, line 107: Will this cout call write to anywhere that a user can see?

ciprianbadescu: This is good question, probably not when running as service, maybe `cerr ` will help when started manually.
Any suggestion for what defense (if any) should we have in case `ReportEvent` is not successful?

McdonaldSeanp: I don't know if there's much we can do when running as a service. I think it's pretty standard for customer support to have folks run pxp-agent manually when something is broken, so I don't think it's a problem that nothing happens if `ReportEvent` fails.

If pxp-agent is running as a service, the fact that events are simply missing from the event log should be enough for someone to recognize that something is wrong. From there they are likely to try running pxp-agent in the foreground (again, this is common practice for debugging) and that's where they will see this error.

Originally I asked the question just to clarify for myself what was going on, it sounds like the answer is "when pxp-agent is running as a service this won't be logged anywhere" and that's fine.

From here I think we should go with you suggestion: change this to `cerr` so it's clear it's a failure when running in the foreground, and then I might add a comment above this line that says this message won't show up when running as a service.

McdonaldSeanp (approved):

Dorin-Pleava (approved):
Tested on Windows 2012 and Windows FIPS, and works as expected.

PR #329, opened 2020-11-24
(PA-3490) Add syslog logging to leatherman

GabrielNagy:


logging/inc/leatherman/logging/logging.hpp
```diff
@@ -162,6 +162,13 @@ namespace leatherman { namespace logging {
      */
     void setup_eventlog_logging(std::string application);
 
+    /**
+     * Configures application to log to syslog
+     * The logging level is set to warning by default.
+     * @param application Destination stream for logging output.
+     */
+    void setup_syslog_logging(std::string application, int facility = (21<<3));
```
GabrielNagy, line 170: should probably have a define for `(21<<3)`

ciprianbadescu: I'm not sure if this parameter should be optional at all

logging/src/logging.cc
```diff
@@ -168,6 +169,8 @@ namespace leatherman { namespace logging {
 
         if (use_event_log) {
             log_eventlog(level, message);
+        } else if (use_syslog) {
+            log_syslog(level, message);
         } else {
```
ciprianbadescu, line 174: maybe we should replace use_event_log/use_event_log flags with logging_backend enum and make the dispatch using switch

ciprianbadescu: we could also move boost related logging to `log_boost`

logging/inc/leatherman/logging/logging.hpp
GabrielNagy: makes sense

logging/src/posix/logging.cc
```diff
@@ -39,4 +40,42 @@ namespace leatherman { namespace logging {
...
+        case log_level::error:
+            return LOG_ERR;
+        case log_level::warning:
+            return LOG_WARNING;
+        case log_level::info:
+            return LOG_INFO;
+        case log_level::debug:
+        case log_level::trace:
+        case log_level::none:
+            return LOG_DEBUG;
```
GabrielNagy, line 69: should not log anything if `level == log_level::none`

ciprianbadescu (approved):

PR #333, opened 2021-12-17
(LTH-172) Add support for RHEL-9(glibc 2.34)

ciprianbadescu:
Adapted from https://github.com/catchorg/Catch2/pull/2317/commits/f83d0328e07243384a7a312a487dde4382600742

On GLIBC >= 2.33, `MINSIGSTKSZ` is not a constant. This can be detected
at compile time by the presence of either `_SC_SIGSTKSZ_SOURCE` or
`_GNU_SOURCE`.

If either are define, hardcode the size we want rather than compare it,
since we can't do that in a constant expression anymore.

The changes applied are:
```
--- a/Catch-1.10.0/single_include/catch.hpp	2021-07-05 15:30:43.000000000 +0300
+++ b/Catch-1.10.0/single_include/catch.hpp	2021-12-17 09:34:41.769565768 +0200
@@ -6447,6 +6447,14 @@
 
 #  else // CATCH_CONFIG_POSIX_SIGNALS is defined
 
+#    if defined(_SC_SIGSTKSZ_SOURCE) || defined(_GNU_SOURCE)
+// on glibc > 2.33 this is no longer constant, see
+// https://sourceware.org/git/?p=glibc.git;a=blob;f=NEWS;h=85e84fe53699fe9e392edffa993612ce08b2954a;hb=HEAD
+static constexpr std::size_t sigStackSize = 32768;
+#    else
+static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
+#    endif
+
 #include <signal.h>
 
 namespace Catch {
@@ -6470,7 +6478,7 @@
         static bool isSet;
         static struct sigaction oldSigActions [sizeof(signalDefs)/sizeof(SignalDefs)];
         static stack_t oldSigStack;
-        static char altStackMem[SIGSTKSZ];
+        static char altStackMem[sigStackSize];
 
         static void handleSignal( int sig ) {
             std::string name = "<unknown signal>";
@@ -6490,7 +6498,7 @@
             isSet = true;
             stack_t sigStack;
             sigStack.ss_sp = altStackMem;
-            sigStack.ss_size = SIGSTKSZ;
+            sigStack.ss_size = sigStackSize;
             sigStack.ss_flags = 0;
             sigaltstack(&sigStack, &oldSigStack);
             struct sigaction sa = { 0 };
@@ -6521,7 +6529,7 @@
     bool FatalConditionHandler::isSet = false;
     struct sigaction FatalConditionHandler::oldSigActions[sizeof(signalDefs)/sizeof(SignalDefs)] = {};
     stack_t FatalConditionHandler::oldSigStack = {};
-    char FatalConditionHandler::altStackMem[SIGSTKSZ] = {};
+    char FatalConditionHandler::altStackMem[sigStackSize] = {};
 
 } // namespace Catch
 
```


joshcooper (approved):
:trophy:

PR #339, opened 2024-01-16
(PE-36763) Migrate CURLOPT_PROTOCOLS{,_STR}

tlehman:
This PR removes the [deprecated `CURLOPT_PROTOCOLS` option](https://curl.se/libcurl/c/CURLOPT_PROTOCOLS.html) and replaces it with the new [`CURLOPT_PROTOCOLS_STR` option](https://curl.se/libcurl/c/CURLOPT_PROTOCOLS_STR.html).

Jira: https://perforce.atlassian.net/browse/PE-36763

# Testing
This builds and the unit tests all pass

DO NOT MERGE UNTIL: all consumers of leatherman (pxp-agent, etc.) have been updated to use this option.

PR #340, opened 2024-02-09
(PE-36763) Add backwards-compat support for CURLOPT_PROTOCOLS_STR

mcdonaldseanp:
Update the leatherman curl client to support usage of CURLOPT_PROTOCOLS_STR when it is available. Versions of cURL earlier than 8 will not have CURLOPT_PROTOCOLS_STR, so these changes allow usage of CURLOPT_PROTOCOLS if the new string version of the function is not available.

The _client_protocols param of the client context is now a string, and there is a second implementation of set_supported_protocols that takes a string in the form CURLOPT_PROTOCOLS_STR expects.

Regardless of which version of cURL leatherman is compiled with, both the bitmask and string versions of set_supported_protocols perform the same logic. However the set_client_protocols function - the function that actually calls CURLOPT_PROTOCOLS or CURLOPT_PROTOCOLS_STR - differs: with older cURL it translates the strings back to bitmasks, with cURL newer than version 8 it sends the string directly to CURLOPT_PROTOCOLS_STR.

Structurally the code was written in an attempt to 1. move everything towards the use of strings with CURLOPT_PROTOCOLS_STR and 2. to avoid differently compiled code between versions as much as possible.

### DIFF
--- a/.github/actions/run_cmake_and_make/action.yaml
+++ b/.github/actions/run_cmake_and_make/action.yaml
@@ -1,4 +1,4 @@
-name: docker_pull_and_make
+name: run_cmake_and_make
 
 inputs:
   pkg_suffix:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.2.2)
 project(leatherman VERSION 1.12.12)
 
+# Set the cmake policy that allows -D<LIB>_ROOT settings, so that you can point to a specific boost or curl
+cmake_policy(SET CMP0074 NEW)
+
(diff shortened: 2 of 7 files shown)


tlehman (approved):
Looks good, closing out [leatherman#339](https://github.com/puppetlabs/leatherman/pull/339/files) in favor of this one

joshcooper:
Getting some build failures in https://jenkins-platform.delivery.puppetlabs.net/view/pxp-agent-vanagon/job/platform_pxp-agent-vanagon_pxp_agent_vanagon-packaging_7.x/152/

```
19:17:05 CMake Error at CMakeLists.txt:5 (cmake_policy):
19:17:05   Policy "CMP0074" is not known to this version of CMake.
```

PR #341, opened 2024-02-09
(PE-37666)(PE-37667) Don't use a docker container for builds

mcdonaldseanp:
We can just build with github's ubuntu-latest images for unit testing


tlehman (approved):
:shipit:

PR #342, opened 2024-02-13
(PE-37481) Open file with SH_DENYNO sharing mode

tlehman:
On Windows, use the `_fsopen` function with the `_SH_DENYNO` sharing mode to allow multiple threads to access the same file.

# NOTE: I am also looking into using the [`CreateFileW`](https://github.com/boostorg/filesystem/issues/305) method mentioned in that issue.

# Testing
I built this branch and ran it as a puppet-agent, and it resolved the error in the ticket, you can see all these tasks succeeded
image

### DIFF
--- a/curl/src/client.cc
+++ b/curl/src/client.cc
@@ -142,6 +142,7 @@ namespace leatherman { namespace curl {
         close_fp();
         boost::system::error_code ec;
         fs::rename(_temp_path, _file_path, ec);
+        LOG_DEBUG("fs::rename {1}, {2}, error_code: {3}", _temp_path, _file_path, ec);
         if (ec) {
             LOG_WARNING("Failed to write the results of the temporary file to the actual file {1}", _file_path);
             throw http_file_operation_exception(_req, _file_path, make_file_err_msg(_("failed to move over the temporary file's downloaded contents")));
--- a/curl/src/client.cc
+++ b/curl/src/client.cc
@@ -10,6 +10,11 @@
 #include <boost/nowide/fstream.hpp>
 #include <sstream>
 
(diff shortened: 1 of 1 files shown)


curl/src/client.cc
```diff
@@ -101,6 +106,9 @@ namespace leatherman { namespace curl {
     {
         try {
             _temp_path = fs::path(file_path).parent_path() / fs::unique_path("temp_file_%%%%-%%%%-%%%%-%%%%");
+            #ifdef _WIN32
+            _fp = _fsopen(_temp_path.string().c_str(), "w+", _SH_DENYNO);
```
donoghuc: This seems to result in an error such that the sha of the downloaded file does not match the expected sha. My guess is that the file is empty, but it could be an encoding issue? I think it may be worth looking in to temporarily keeping around the tempfile to take a look at what it looks like. Trace level debugging may help also.

donoghuc (changes requested):

curl/src/client.cc
```diff
@@ -107,7 +107,7 @@ namespace leatherman { namespace curl {
         try {
             _temp_path = fs::path(file_path).parent_path() / fs::unique_path("temp_file_%%%%-%%%%-%%%%-%%%%");
             #ifdef _WIN32
-            _fp = _fsopen(_temp_path.string().c_str(), "w+", _SH_DENYNO);
+            _fp = _fsopen(_temp_path.string().c_str(), "wb", _SH_DENYNO);
```
donoghuc, line 110: Ah, i bet that is it. Probably the line endings were encoded as `\r\n` without this. Good catch.

PR #345, opened 2024-02-20
(PE-37481) Add retry with exponential backoff when error 32 (file in use on windows) occurs

tlehman:
This PR adds:

- Instructions for running the tests
- A retry loop with exponential backoff (100ms up to 5 times, doubling each time) when there's a system error 32
    - This is to avoid having to make windows-specific code

### DIFF
--- a/curl/src/client.cc
+++ b/curl/src/client.cc
@@ -137,7 +137,7 @@ namespace leatherman { namespace curl {
     }
 
     bool error_file_in_use(const boost::system::error_code& ec) {
-        bool same = (ec == boost::system::error_code(32, boost::system::system_category()));
+        bool same = (ec.value() == 32);
         LOG_DEBUG("error_file_in_use({1}), same = {2}", ec, same);
         return same;
     }
@@ -153,7 +153,7 @@ namespace leatherman { namespace curl {
             fs::rename(_temp_path, _file_path, ec);
             wait_millis *= 2;
             max_retries -= 1;
-        } while (max_retries > 0 && error_file_in_use(ec));
+        } while (error_file_in_use(ec) && max_retries > 0);
     }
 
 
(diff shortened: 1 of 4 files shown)


curl/src/client.cc
```diff
@@ -8,6 +8,8 @@
 #include <boost/algorithm/string.hpp>
 #include <boost/nowide/iostream.hpp>
 #include <boost/nowide/fstream.hpp>
+#include <boost/chrono.hpp>
```
donoghuc, line 11: Are you sure these are correct? It looks like it should be 
```
#include <boost/chrono/chrono.hpp>
#include <boost/thread/thread.hpp>
```
Check out how we include these libs in cpp-pcp-client. Also, you can see how we use those in pxp-agent. 
https://github.com/puppetlabs/cpp-pcp-client/blob/b22db78a97a78e7f8cc189223af820756d4c0a8c/lib/inc/cpp-pcp-client/util/chrono.hpp#L4C1-L4C35
https://github.com/puppetlabs/cpp-pcp-client/blob/b22db78a97a78e7f8cc189223af820756d4c0a8c/lib/inc/cpp-pcp-client/util/thread.hpp#L7C1-L7C35

tlehman: It works on macOS, one sec

tlehman: Yep, builds both ways on macOS, going to try this on Windows

donoghuc: We use this exact pattern (sleeping in a thread using boost thread and chrono) in pxp-agent. its just done through cpp-pcp-client util. I would just compare exactly what is done there and how it works for that project. https://github.com/puppetlabs/pxp-agent/blob/276a8ac0ef8b6e36564cdcf78da2717b1865be35/lib/src/request_processor.cc#L548-L549

donoghuc: Having a look at the warnings in the appveyor tests (not sure what is broken in the ubuntu ones) it looks like https://github.com/puppetlabs/cpp-pcp-client/commit/f478d7dcc4e41deae63d1de1bf777a2db0c7617a to me. I would have a look through that header file for things we may want to add here.

donoghuc: Another commit in that file that looks suspicious https://github.com/puppetlabs/cpp-pcp-client/commit/c67710c637aca9499196d33ed7164d22b326b49c

tlehman:
Closing this in favor of a retry loop at the pxp-agent level. This logic doesn't belong in leatherman.

PR #346, opened 2024-02-21
(PE-36763) Add backwards-compat support for CURLOPT_PROTOCOLS_STR

mcdonaldseanp:
Update the leatherman curl client to support usage of CURLOPT_PROTOCOLS_STR when it is available. Versions of cURL earlier than 8 will not have CURLOPT_PROTOCOLS_STR, so these changes allow usage of CURLOPT_PROTOCOLS if the new string version of the function is not available.

The _client_protocols param of the client context is now a string, and there is a second implementation of set_supported_protocols that takes a string in the form CURLOPT_PROTOCOLS_STR expects.

Regardless of which version of cURL leatherman is compiled with, both the bitmask and string versions of set_supported_protocols perform the same logic. However the set_client_protocols function - the function that actually calls CURLOPT_PROTOCOLS or CURLOPT_PROTOCOLS_STR - differs: with older cURL it translates the strings back to bitmasks, with cURL newer than version 8 it sends the string directly to CURLOPT_PROTOCOLS_STR.

Structurally the code was written in an attempt to 1. move everything towards the use of strings with CURLOPT_PROTOCOLS_STR and 2. to avoid differently compiled code between versions as much as possible.

### DIFF
--- a/curl/tests/client_test.cc
+++ b/curl/tests/client_test.cc
@@ -34,7 +34,7 @@ fs::path find_matching_file(const boost::regex& re) {
         fs::recursive_directory_iterator(fs::current_path()),
         fs::recursive_directory_iterator(),
         [re](const fs::path& f) { return boost::regex_match(f.filename().string(), re); });
-  
+
     // throw exception, as this means that the matching file does not exist.
     if (file == fs::recursive_directory_iterator()) {
       throw std::runtime_error("matching file not found");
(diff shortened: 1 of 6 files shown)


curl/tests/client_test.cc
```diff
@@ -297,18 +297,43 @@ TEST_CASE("curl::client CA bundle and SSL setup") {
...
+        REQUIRE(test_impl->protocols == "http");
+    }
+
+    SECTION("cURL should make an HTTP request with multiple specified protocols") {
+        test_client.set_supported_protocols("http,https");
+        auto resp = test_client.get(test_request);
+        CURL* const& handle = test_client.get_handle();
+        auto test_impl = reinterpret_cast<curl_impl* const>(handle);
+        // Either order is fine
+        REQUIRE((test_impl->protocols == "http,https" || test_impl->protocols == "https,http"));
```
tlehman, line 313: This could be collapsed by using an `bool operator==` overload and then making `given_protocols` into a `std::set`. But it's probably overkill. I just love overloading operators.

tlehman (approved):
:shipit:

donoghuc (approved):

